PR #2, opened 2019-03-07
Building_area indicator + dataset for tests

j3r3m1:
I have coded the building_area indicator and added a sql file used for the indicator tests.


ebocher:
I'm on it

ebocher:
Should be organized according the PR #4

PR #5, opened 2019-03-20
Pom updates

SPalominos:
Fixes the pom in order to use gmavenplus instead of gmaven plugin.
@j3r3m1 It should solve your issue.


j3r3m1:
yes I confirm I have run your master it does work with your modifications.

SPalominos:
Great! I merge it.

PR #6, opened 2019-03-20
Building size

j3r3m1:
Some of the indicators used for building size properties are calculated. I could not add the "building_likelihood_large_building" indicator since it needs the "building_nb_neighboor" indicator value as input.


geoindicators/src/test/groovy/org/orbisgis/BuildingIndicatorsTests.groovy
```diff
@@ -26,4 +26,27 @@ class BuildingIndicatorsTests {
         }
     }
 
+    @Test
+    void buildingSizeProperties() {
+        def h2GIS = H2GIS.open([databaseName: './target/buildingdb'])
+        String sqlFilePath = System.getProperty("user.dir")+"/src/test/resources/org/orbisgis/data_for_tests.sql"
```
ebocher: Why not use getClass().getResourceAsStream

j3r3m1: I would love but I do not know how to use it. I have tried 
`this.getClass().getResourceAsStream("../data_for_tests.sql")`
but it does now work very well...

ebocher: @SPalominos

SPalominos: You can use 
 ``` groovy
String sqlString = new File(this.class.getResource("data_for_tests.sql").toURI()).text
```

ebocher:
Excellent !

PR #7, opened 2019-03-21
Coded the buildingNeighborsProperties

j3r3m1:
The building_road_distance and the building_minimum_building_spacing will be coded each in a separate IProcess.

### DIFF
--- a/geoindicators/src/main/groovy/org/orbisgis/BuildingIndicators.groovy
+++ b/geoindicators/src/main/groovy/org/orbisgis/BuildingIndicators.groovy
@@ -101,8 +101,8 @@ static IProcess buildingNeighborsProperties() {
                 def build_intersec = "build_intersec"+uid_out.toString()
 
 
-                String query = "CREATE SPATIAL INDEX IF NOT EXISTS buff_ids ON $inputBuildingTableName($geometricField) USING rtree; " +
-                        "CREATE INDEX IF NOT EXISTS buff_id ON $inputBuildingTableName($idField) USING rtree;" +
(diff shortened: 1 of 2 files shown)


geoindicators/src/main/groovy/org/orbisgis/BuildingIndicators.groovy
```diff
@@ -76,3 +76,72 @@ static IProcess buildingSizeProperties() {
...
+                def ops = ["building_contiguity","building_common_wall_fraction",
+                           "building_number_building_neighbor"]
+                // To avoid overwriting the output files of this step, a unique identifier is created
+                def uid_out = System.currentTimeMillis()
+                // Temporary table names
+                def build_intersec = "build_intersec"+uid_out.toString()
+
+
+                String query = "CREATE SPATIAL INDEX IF NOT EXISTS buff_ids ON $inputBuildingTableName($geometricField); " +
+                        "CREATE INDEX IF NOT EXISTS buff_id ON $inputBuildingTableName($idField);" +
```
ebocher: Please use
CREATE INDEX IF NOT EXISTS buff_ids ON $inputBuildingTableName($geometricField) USING rtree;

j3r3m1: I have tried to implement your piece of code for all my index creation but it does not work. I got the following error:
`Syntax error in SQL statement {0}; expected {1}; SQL statement:`

ebocher:
Thanks

PR #11, opened 2019-03-25
buildingFormProperties coded.

j3r3m1:


geoindicators/src/main/groovy/org/orbisgis/BuildingIndicators.groovy
```diff
@@ -140,6 +140,46 @@ static IProcess buildingNeighborsProperties() {
...
+static IProcess buildingFormProperties() {
+    return processFactory.create(
+            "Building form properties",
+            [inputBuildingTableName: String,inputFields:String[],operations: String[]
+             , outputTableName: String, datasource: JdbcDataSource],
+            [outputTableName : String],
+            { inputBuildingTableName,inputFields, operations, outputTableName, datasource ->
+                def geometricField = "the_geom"
+                def height_wall = "height_wall"
+                def ops = ["building_concavity","building_form_factor",
```
ebocher: name it without the building word prefix because it could be used for blocks ?

j3r3m1: We do not calculate concavity neither form factor for blocks. It could be interesting anyway. But it would imply to remove the raw_compacity from the list since at block scale, the definition is different (the roof height of each building being considered for the calculation).

ebocher: good catch

PR #13, opened 2019-03-26
Building spacing

j3r3m1:
coded the building_minimum_building_spacing indicator

### DIFF
--- a/geoindicators/src/main/groovy/org/orbisgis/BuildingIndicators.groovy
+++ b/geoindicators/src/main/groovy/org/orbisgis/BuildingIndicators.groovy
@@ -177,7 +177,9 @@ static IProcess buildingNeighborsProperties() {
  * --> "building_form_factor": defined as ratio between the building area divided by the square of the building
  * perimeter (cf. Bocher et al. - 2018)
  * --> "building_raw_compacity": defined as the ratio between building surfaces (walls and roof) divided by the
- * building volume at the power 2./3.
+ * building volume at the power 2./3. For the calculation, the roof is supposed to have a gable and the roof surface
+ * is calculated considering that the building is square (otherwise, the assumption related to the gable direction
+ * would strongly affect the result).
  *
  * References:
  *   Bocher, E., Petit, G., Bernard, J., & Palominos, S. (2018). A geoprocessing framework to compute
@@ -195,6 +197,7 @@ static IProcess buildingFormProperties() {
             { inputBuildingTableName,inputFields, operations, outputTableName, datasource ->
                 def geometricField = "the_geom"
                 def height_wall = "height_wall"
+                def height_roof = "height_roof"
                 def ops = ["building_concavity","building_form_factor",
                            "building_raw_compacity"]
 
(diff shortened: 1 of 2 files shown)


geoindicators/src/main/groovy/org/orbisgis/BuildingIndicators.groovy
```diff
@@ -144,4 +167,116 @@ static IProcess buildingNeighborsProperties() {
...
+                                    "CREATE INDEX IF NOT EXISTS with_buff_id ON $build_within_buffer($idField);"+
+                                    "CREATE INDEX IF NOT EXISTS with_buff_ids ON $build_within_buffer($geometricField) "+
+                                    "USING RTREE").toString())
+                // The minimum distance is calculated
+                datasource.execute(("DROP TABLE IF EXISTS $outputTableName; CREATE TABLE $outputTableName AS "+
+                                    "SELECT MIN(ST_DISTANCE(a.$geometricField, b.$geometricField)) AS building_minimum_building_spacing, "+
+                                    "a.${inputFields.join(",a.")} FROM $build_within_buffer b RIGHT JOIN $inputBuildingTableName a ON a.$idField = b.$idField"+
+                                    " GROUP BY a.$idField").toString())
+                // The minimum distance is set to the $inputE value for buildings having no building neighbors in a $inputE meters distance
+                datasource.execute(("UPDATE $outputTableName SET building_minimum_building_spacing = $bufferDist "+
```
ebocher: To avoid an update command you can set a default value when the table is created.

 CREATE TABLE $outputTableName(building_minimum_building_spacing float DEFAULT  $bufferDist...) as select

geoindicators/src/main/groovy/org/orbisgis/BuildingIndicators.groovy
```diff
@@ -144,4 +167,116 @@ static IProcess buildingNeighborsProperties() {
...
+ * This process extract the building closest distance to an other building. A buffer of defined size (bufferDist argument)
+ * is used to get the buildings within the building of interest and then the minimum distance is calculated.
+ *
+ * @return A database table name.
+ * @author Jérémy Bernard
+ */
+static IProcess buildingMinimumBuildingSpacing() {
+    return processFactory.create(
+            "Building minimum building spacing",
+            [inputBuildingTableName: String,inputFields:String[],bufferDist: Integer
```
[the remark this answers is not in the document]
j3r3m1: ok

geoindicators/src/main/groovy/org/orbisgis/BuildingIndicators.groovy
j3r3m1: I will update that =)

j3r3m1: The problem is that if I use your query I will have to define the type of the inputFields since I call them in the `CREATE TABLE $outputTableName` query. We do not have this information in advance. Is there a way to get the type of a field within a given Table ? Something like:
`datasource.getTable(inputBuildingTableName).getField(fieldName)`

PR #14, opened 2019-03-26
Building road dist

j3r3m1:


j3r3m1:
About this indicator and the building distance one, will it be possible to set a default value for some of the input parameters used in the IProcess ? For these specific IProcesses, the bufferDist param should be set by default to 100 m.

ebocher:
You can initialize a distance with a default value in the closure. Check if the input distance is greater than the default value or less than a specific value then use the default one.

j3r3m1:
OK, I have set the default value. I do not think it is necessary to define any condition (lower than the default), do you ?

geoindicators/src/main/groovy/org/orbisgis/BuildingIndicators.groovy
```diff
@@ -239,7 +239,7 @@ static IProcess buildingMinimumBuildingSpacing() {
             [inputBuildingTableName: String,inputFields:String[],bufferDist: Double
              , outputTableName: String, datasource: JdbcDataSource],
             [outputTableName : String],
-            { inputBuildingTableName,inputFields, bufferDist, outputTableName, JdbcDataSource datasource ->
+            { inputBuildingTableName,inputFields, bufferDist = 100, outputTableName, datasource ->
```
ebocher: +1

ebocher:
Brilliant

PR #17, opened 2019-03-26
Operation from lower scale

j3r3m1:
Code for basic statistical operations (sum and average) from a scale to another (building to block, building to rsu or block to rsu).

Useful for indicators such as: block_building_volume, block_building_area, block_floor_area, etc.

Another indicator will be used for weighted statistical operations.

### DIFF
--- a/geoindicators/src/main/groovy/org/orbisgis/BlockIndicators.groovy
+++ b/geoindicators/src/main/groovy/org/orbisgis/BlockIndicators.groovy
@@ -11,7 +11,7 @@ import org.orbisgis.processmanagerapi.IProcess
  * This process is used to compute basic statistical operations on a specific variable from a lower scale (for
  * example the sum of each building volume constituting a block to calculate the block volume)
  * @return A database table name.
- * @author Erwan Bocher
+ * @author Jérémy Bernard
  */
 static IProcess unweightedOperationFromLowerScale() {
 return processFactory.create(
--- a/geoindicators/src/main/groovy/org/orbisgis/BlockIndicators.groovy
+++ b/geoindicators/src/main/groovy/org/orbisgis/BlockIndicators.groovy
@@ -1,0 +1,42 @@
+package org.orbisgis
+
+import groovy.transform.BaseScript
+import org.orbisgis.datamanager.JdbcDataSource
+import org.orbisgis.datamanagerapi.dataset.ITable
+import org.orbisgis.processmanagerapi.IProcess
+
+@BaseScript Geoclimate geoclimate
+
+/**
+ * This process is used to compute basic statistical operations on a specific variable from a lower scale (for
+ * example the sum of each building volume constituting a block to calculate the block volume)
+ * @return A database table name.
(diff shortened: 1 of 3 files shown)


geoindicators/src/main/groovy/org/orbisgis/BlockIndicators.groovy
```diff
@@ -0,0 +1,42 @@
...
+
+@BaseScript Geoclimate geoclimate
+
+/**
+ * This process is used to compute basic statistical operations on a specific variable from a lower scale (for
+ * example the sum of each building volume constituting a block to calculate the block volume)
+ * @return A database table name.
+ * @author Erwan Bocher
+ */
+static IProcess unweightedOperationFromLowerScale() {
```
ebocher: aggregatedStatistics ?

geoindicators/src/main/groovy/org/orbisgis/BlockIndicators.groovy
```diff
@@ -0,0 +1,42 @@
...
+return processFactory.create(
+        "Unweighted statistical operations from lower scale",
+        [inputLowerScaleTableName: String,inputCorrelationTableName: String,inputIdLow: String, inputIdUp: String,
+         inputToTransfo: String,operations: String[], outputTableName: String, datasource: JdbcDataSource],
+        [outputTableName : String],
+        { inputLowerScaleTableName, inputCorrelationTableName, inputIdLow, inputIdUp,
+          inputToTransfo, operations, outputTableName, datasource ->
+            def ops = ["SUM","AVG"]
```
ebocher: count, min, max ???

geoindicators/src/main/groovy/org/orbisgis/BlockIndicators.groovy
j3r3m1: unweightedAggregatedStatistics in order to differentiate from the weighted ones ?

geoindicators/src/main/groovy/org/orbisgis/BlockIndicators.groovy
j3r3m1: So far (for the indicators we have defined) there is no need for such functions. However, we could add them if you think it is necessary for the future.

geoindicators/src/main/groovy/org/orbisgis/BuildingIndicators.groovy
```diff
@@ -222,6 +222,116 @@ static IProcess buildingFormProperties() {
...
+ * This process extract the building closest distance to an other building. A buffer of defined size (bufferDist argument)
+ * is used to get the buildings within the building of interest and then the minimum distance is calculated.
+ *
+ * @return A database table name.
+ * @author Jérémy Bernard
+ */
+static IProcess buildingMinimumBuildingSpacing() {
+    return processFactory.create(
+            "Building minimum building spacing",
+            [inputBuildingTableName: String,inputFields:String[],bufferDist: Integer
```
ebocher: double ?

PR #20, opened 2019-03-27
Block direction

j3r3m1:


geoindicators/src/main/groovy/org/orbisgis/BlockIndicators.groovy
```diff
@@ -0,0 +1,213 @@
...
+                def idFieldBu = "id_build"
+                def idFieldBl = "id_block"
+
+                // Test whether the angleRangeSize is a divisor of 180°
+                if (180%angleRangeSize==0 & 180/angleRangeSize>1){
+                    // To avoid overwriting the output files of this step, a unique identifier is created
+                    def uid_out = System.currentTimeMillis()
+
+                    // Temporary table names
+                    def build_min_rec = "build_min_rec"+uid_out.toString()
```
ebocher: "build_min_rec"+uid_out should be enough

PR #22, opened 2019-03-27
Coded rsuFreeExternalFacadeDensity + modification of the test dataset

j3r3m1:


geoindicators/src/main/groovy/org/orbisgis/RsuIndicators.groovy
```diff
@@ -0,0 +1,50 @@
...
+ * the "building_total_facade_length" values as well as a correlation Table between buildings and blocks.
+ *
+ * @return A database table name.
+ * @author Jérémy Bernard
+ */
+static IProcess rsuFreeExternalFacadeDensity() {
+return processFactory.create(
+        "RSU free external facade density",
+        [inputBuildingTableName: String,inputFields:String[],inputCorrelationTableName: String,
+         buContiguityFieldName: String, buTotalFacadeLength: String, outputTableName: String, datasource: JdbcDataSource],
```
ebocher: buTotalFacadeLength - > field name ?
buContiguityFieldName -> inputFieldName ?

j3r3m1: Both are input field names (we should have calculated the buildingTotalFacadeLength and the buildingContiguity to calculate the rsuFreeExternalFacadeDensity - at least it is the fastest solution). If I only described them as inputFieldName we can not differentiate them.

 Is buTotalFacadeLengthFieldName and buContiguityFieldName OK ?

ebocher:
To be merged

PR #23, opened 2019-03-28
Rsu svf

j3r3m1:


geoindicators/src/main/groovy/org/orbisgis/RsuIndicators.groovy
```diff
@@ -0,0 +1,162 @@
...
+ * @author Jérémy Bernard
+ */
+static IProcess rsuGroundSkyViewFactor() {
+    return processFactory.create(
+            "RSU ground sky view factor",
+            [rsuTable: String,inputColumns:String[],correlationBuildingTable: String,
+             rsuAreaColumn: String, rsuBuildingDensityColumn: String, pointDensity: Double, rayLength: Double,
+             numberOfDirection: Double, outputTableName: String, datasource: JdbcDataSource],
+            [outputTableName : String],
+            { rsuTable, inputColumns, correlationBuildingTable, rsuAreaColumn, rsuBuildingDensityColumn,
```
j3r3m1: I did not consider the inputColumns (potential columns that the user would like to keep) in the code.
Should we delete this inpuData ?

ebocher: Keep only an id field will be enough for me

PR #41, opened 2019-04-04
First processing chain for BDTopo

ebocher:
Good luck :-)


pom.xml
```diff
@@ -15,6 +15,7 @@
     <modules>
         <module>geoindicators</module>
         <module>prepare_data</module>
+	<module>processing_chain</module>
```
SPalominos: The module artifact name should be `processing-chain` with a `-`

processing_chain/pom.xml
```diff
@@ -0,0 +1,98 @@
...
+    <modelVersion>4.0.0</modelVersion>
+
+    <properties>
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+        <maven.build.timestamp.format>YYYY-MM-dd</maven.build.timestamp.format>
+        <buildNumber>${maven.build.timestamp}</buildNumber>
+        <site.basedir>${project.parent.basedir}</site.basedir>
+    </properties>
+
+    <artifactId>processing_chain</artifactId>
```
SPalominos: The module artifact name should be processing-chain with a -

processing_chain/pom.xml
```diff
@@ -0,0 +1,98 @@
...
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-simple</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.orbisgis</groupId>
+            <artifactId>prepare_data</artifactId>
+            <version>1.0-SNAPSHOT</version>
+            <scope>compile</scope>
+        </dependency>
```
SPalominos: The prepare_data dependency should be before the `<!-- Test dependencies -->` tag (line 41)

processing_chain/src/main/groovy/org/orbisgis/processingchain/PrepareBDTopo.groovy
```diff
@@ -0,0 +1,26 @@
...
+import org.orbisgis.common.AbstractTablesInitialization
+import org.orbisgis.processmanager.ProcessMapper
+import org.orbisgis.processmanagerapi.IProcess
+
+@BaseScript ProcessingChain processingChain
+
+/**
+ *
+ * @return
+ */
```
SPalominos: The javadoc is empty

processing_chain/src/test/groovy/org/orbisgis/processingchain/ProcessingChainTest.groovy
```diff
@@ -0,0 +1,30 @@
+package org.orbisgis.processingchain
+
+import org.junit.jupiter.api.Test
+
+import org.orbisgis.datamanager.h2gis.H2GIS
+import org.orbisgis.processmanager.ProcessMapper
+
+class ProcessingChainTest {
+
```
SPalominos: In order to skip this test when the resources are not present (as the geoclimate_bdtopo_data_test should not be published) add the following method which register a system property which will be used to disable test if needed.
``` groovy
@BeforeAll
static void init(){
    System.setProperty("test.processingchain",
            Boolean.toString(ProcessingChainTest.getResource("geoclimate_bdtopo_data_test") != null))
}
```

processing_chain/src/test/groovy/org/orbisgis/processingchain/ProcessingChainTest.groovy
```diff
@@ -0,0 +1,30 @@
...
+    @Test
+    void runBDTopoProcessingChain(){
```
SPalominos: Thanks to the method from the previous comment, you can add :
``` groovy
@EnabledIfSystemProperty(named = "test.processingchain", matches = "true")
```
to disable the test if the resources are not present.

processing_chain/src/test/groovy/org/orbisgis/processingchain/ProcessingChainTest.groovy
```diff
@@ -0,0 +1,30 @@
...
+        h2GIS.load(new File(this.class.getResource("geoclimate_bdtopo_data_test/TRONCON_VOIE_FERREE.geojson").toURI()).getAbsolutePath(),"TRONCON_VOIE_FERREE",true)
+        h2GIS.load(new File(this.class.getResource("geoclimate_bdtopo_data_test/SURFACE_EAU.geojson").toURI()).getAbsolutePath(),"SURFACE_EAU",true)
+        h2GIS.load(new File(this.class.getResource("geoclimate_bdtopo_data_test/ZONE_VEGETATION.geojson").toURI()).getAbsolutePath(),"ZONE_VEGETATION",true)
+
+        ProcessMapper pm =  ProcessingChain.prepareBDTopo.createMapper()
+        pm.execute([datasource: h2GIS, distBuffer : 500, expand : 1000, idZone : "56260", tableIrisName: "IRIS_GE",
+                    tableBuildIndifName: "BATI_INDIFFERENCIE", tableBuildIndusName: "BATI_INDUSTRIEL",
+                    tableBuildRemarqName: "BATI_REMARQUABLE", tableRoadName: "ROUTE", tableRailName: "TRONCON_VOIE_FERREE",
+                    tableHydroName: "SURFACE_EAU", tableVegetName: "ZONE_VEGETATION"])
+
```
SPalominos: Add at least an assertion of the test.

pom.xml
j3r3m1: done

processing_chain/pom.xml
j3r3m1: done

processing_chain/pom.xml
j3r3m1: done

processing_chain/src/main/groovy/org/orbisgis/processingchain/PrepareBDTopo.groovy
j3r3m1: I fill it out. However, I wonder if the @param and @return parameters should be filled here. It would need to get into each of the processes to copy and paste the same name and definition. Would it be possible to get them automatically or should we copy and paste each of them (if we have several mappers using the same process, a modification of the process would need to modify all the mappers using it...)

SPalominos: It would be great but quite tricky. I keep this idea in mind for a future improvement.

ebocher: Do not copy paste all parameters for each processes. Just describe the required parameters for the whole chain.

pom.xml
j3r3m1: then I have renamed the module and folder name as well. But I wonder why we modify the processing_chain name but keep the prepare_data (why don't we replace it by prepare-data ?)

processing_chain/src/test/groovy/org/orbisgis/processingchain/ProcessingChainTest.groovy
j3r3m1: done I have added the same tests that you did on the last process of the mapper (test whether all output tables have been created in the datasource)

processing_chain/src/test/groovy/org/orbisgis/processingchain/ProcessingChainTest.groovy
j3r3m1: I add it right before the @test line since I had a flag if I put it right after the void (inside the {} of the test). It seems to work like that...

pom.xml
SPalominos: It would be replaced later as @ELSW56 was working on it.

PR #43, opened 2019-04-05
Import osm data

ELSW56:
Creation of the process to import the data from OSM with the overpass API and to build the first tables (ZONE, ZONE_EXTENDED, ZONE_BUFFER and ZONE_NEIGHBORS). The corresponding tests are also released.
In OSMGISLayers, possibility to run the complete process.

### DIFF
--- a/prepare_data/src/main/groovy/org/orbisgis/osm/OSMGISLayers.groovy
+++ b/prepare_data/src/main/groovy/org/orbisgis/osm/OSMGISLayers.groovy
@@ -230,6 +230,78 @@ static IProcess prepareHydro() {
     )
 }
 
+/**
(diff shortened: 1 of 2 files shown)


prepare_data/src/main/groovy/org/orbisgis/osm/OSMGISLayers.groovy
```diff
@@ -230,6 +230,78 @@ static IProcess prepareHydro() {
     )
 }
 
+/**
+ * This process is used to create the hydro table thanks to the osm data tables
```
ebocher: hydro table ?

ELSW56: Fixed !

PR #53, opened 2019-04-11
The createScalesRelations IProcess has been coded and the test work +…

j3r3m1:
… the dataset for the tests has been modified


geoindicators/src/main/groovy/org/orbisgis/SpatialUnits.groovy
```diff
@@ -185,3 +187,40 @@ static IProcess createBlocks(){
             }
     )
 }
+
+/**
+ * This process is used to link each object of a lower scale to an upper scale ID (i.e.: building to RSU)
```
ebocher: maybe add information about the method. Because we use the max intersection area?

j3r3m1: you are right I do that.

PR #56, opened 2019-04-15
Modify nonWeightedAgregated... in order to make it work for several i…

j3r3m1:
…nformations (use of a map) and add the density as operation


geoindicators/src/main/groovy/org/orbisgis/BlockIndicators.groovy
```diff
@@ -11,36 +11,70 @@ import org.orbisgis.processmanagerapi.IProcess
...
-                    if(operation=="GEOM_AVG"){
-                        query += "EXP(1.0/COUNT(*)*SUM(LOG(a.$inputToTransfo))) AS ${operation+"_"+inputToTransfo},"
-                    }
-                    else{
-                        query += "$operation(a.$inputToTransfo::float) AS ${operation+"_"+inputToTransfo},"
+
+            inputVarAndOperations.each {var, operations ->
+                operations.each{op ->
+                    if(ops.contains(op)){
+                        if(op=="GEOM_AVG"){
```
ebocher: use op.toUpperCase() to force the case so user is able to write sum, Sum...

PR #59, opened 2019-04-16
vegetation and road density have been coded, just need to wait for th…

j3r3m1:
…e validation of rsuRoadOperations pull request where there are new data in the dataset that are useful for the test


j3r3m1:
Not able to merge now since I need the validation of some other pull request to use some new data that I have inserted in the dataset for test file.

PR #64, opened 2019-04-30
Add new blockCompacity indicator and modified the building indic used…

j3r3m1:
… for building_volume and building_contiguity indic


ebocher:
Seems a pb with a unit test

j3r3m1:
Yes the one of the hole area but I can not figure it out...

j3r3m1:
The problem remains, I do no understand where does it come from. The test when run alone works but it fails when I run the mvn clean install or sometimes (but only sometimes) the entire test file... @SPalominos  any idea ?

j3r3m1:
OK I think I got the problem... Regarding the modifications I have made, two assumptions, both a bit strange:
- either the problem came from the order of the test regarding the process order in the process file. The order was not conserved which would make the tests failed...
- the string that defined the IProcess (at the beginning of each IProcess) was the same for holeAreaDensity and netCompacity, that could affect the good working of the code...

PR #66, opened 2019-05-02
Coded the pervious and impervious fraction in a same IProcess...

j3r3m1:


j3r3m1:
I could not separate the two IProcesses since they were very simple (simply addition)

PR #67, opened 2019-05-03
Add processing chain createScaleOfAnalysis

j3r3m1:
It can not work now for several reasons:
- a same IProcess (scalesRelations) is used several times in the mapper but does not used the same name each time for the input called columnIdUp. @SPalominos  @ebocher How could we manage this problem (use a same input twice but without the same value...) ?
- the prefix_name is not used in all processes and may disturb the results. I will have to modify all occurences before to run the tests
- the tests are not entirely coded yet since I need the answer to my first problem...


processing-chain/src/main/groovy/org/orbisgis/processingchain/CreateScalesOfAnalysis.groovy
```diff
@@ -0,0 +1,64 @@
...
+ *
+ * @param zoneTable The area of zone to be processed
+ * @param roadTable The road table to be processed
+ * @param railTable The rail table to be processed
+ * @param vegetationTable The vegetation table to be processed
+ * @param hydrographicTable The hydrographic table to be processed
+ * @param surface_vegetation The minimum area of vegetation that will be considered to delineate the RSU (default 100,000 m²)
+ * @param surface_hydro  The minimum area of water that will be considered to delineate the RSU (default 2,500 m²)
+ * @param inputTableName The input table where are stored the geometries used to create the block (e.g. buildings...)
+ * @param distance A distance to group two geometries (e.g. two buildings in a block - default 0.01 m)
```
ebocher: There is any distance parameter in the mapper ?

j3r3m1: Yes, actually we use this parameter as input of the mapper since it is an input of the "createBlocks" IProcess. If two buildings are very close, they are considered as neighbors and they are merged. But I am actually not sure it is a good idea to set 0.01 m as default value since some of the indicators use the merged buildings created on the fly in their own IProcess. Then it would lead to a difference of geometries used as input between the IProcesses. May be setting to 0 in the "createBlocks" IProcess for now is the right solution and we will see later if we get a problem with geometries.

ebocher: +100

PR #72, opened 2019-05-09
Mapper create spatial analysis

j3r3m1:
The mapper has been updated but several problems have been identified:
- the tests of the test file are not run...
- when we get the results of a mapper, are they organized by IProcess such as:
```groovy
mapper.getResults().iprocess1.output1
```
If not, how could we manage to get output1 if I have a second called "iprocess2" that has the same output name "output1" ?


geoindicators/src/main/groovy/org/orbisgis/SpatialUnits.groovy
```diff
@@ -14,25 +14,27 @@ import org.orbisgis.processmanagerapi.IProcess
...
  * @param datasource A connection to a database
  *
- * @return A database table name.
+ * @return A database table name and the name of the column ID
  */
 static IProcess createRSU(){
     return processFactory.create("Create reference spatial units (RSU)",
             [inputTableName: String, prefixName: String, datasource: JdbcDataSource],
-            [outputTableName : String],
+            [outputTableName : String, outputIdRsu: String],
```
ebocher: don't understand why the id column is an output value

geoindicators/src/main/groovy/org/orbisgis/SpatialUnits.groovy
```diff
@@ -177,25 +179,27 @@ static IProcess prepareRSUData(){
...
                 String outputTableName = prefixName + "_" + baseName
 
                 datasource.execute "DROP TABLE IF EXISTS $outputTableName".toString()
-                datasource.execute "CREATE TABLE $outputTableName as  select  EXPLOD_ID as id_block, the_geom ".toString()+
+                datasource.execute "CREATE TABLE $outputTableName as  select  EXPLOD_ID as $columnIdName, the_geom ".toString()+
                         "from st_explode ('(select ST_UNION(ST_ACCUM(ST_BUFFER(THE_GEOM,$distance))) as the_geom".toString()+
                         " from $inputTableName)')".toString()
                 logger.info("The geometries have been merged")
-                [outputTableName: outputTableName]
+                [outputTableName: outputTableName, outputIdBlock: columnIdName]
```
ebocher: idem.
You are able to obtain it with the table metadata ?

geoindicators/src/main/groovy/org/orbisgis/SpatialUnits.groovy
j3r3m1: I added it in order to map it as an input (idColumnUp) in the createScalesRelations IProcess

geoindicators/src/main/groovy/org/orbisgis/SpatialUnits.groovy
j3r3m1: may be but I need to map it as explained above

geoindicators/src/main/groovy/org/orbisgis/SpatialUnits.groovy
ebocher: id column is a constant field so you can pass it with the same value id_rsu, id_block
Am I wrong ?

j3r3m1: yes but a IProcess could be used several times in a mapper and then could have a different name. Then it is necessary to map it to know where does the id used as input come from (rsu or block scale)

ebocher: ok good catch

PR #77, opened 2019-05-09
Modify the location of three IProcesses that was generic instead of p…

j3r3m1:
…roper to a specific scale


j3r3m1:
All tests passed but there is a problem with a one of the processing chain that I have coded. I think I would need to know which IProcess is being displayed in order to easier debug the problem. Thus I will wait for the PR of @ebocher concerning this point.

j3r3m1:
I still have a problem with the mapper createScalesOfAnalysis.
I have a null table in the IProcess createScalesRelations between blocks and buildings. It seems that the mapper input inputLowerScaleTableName is null whereas it is well written as a string in input in my test.
Is this problem could come from the fact that I use several times the createScalesRelations and then the input is defined several times ? I do not get it. @SPalominos I am available for debugging if you have time.

PR #82, opened 2019-05-15
Create new processing chains

j3r3m1:


processing_chain/src/main/groovy/org/orbisgis/processingchain/ProcessingChains.groovy
```diff
@@ -0,0 +1,8 @@
+package org.orbisgis.processingchain
+
+import groovy.transform.BaseScript
+import org.orbisgis.SpatialUnits
+import org.orbisgis.processmanager.ProcessMapper
+
+@BaseScript ProcessingChain processingChain
```
ebocher: not necessary

PR #84, opened 2019-05-15
CreateUnit coded and work

j3r3m1:
@ebocher I did not use your syntax for the beginning of the script because it did not work and I could not figure out why. Let's investigate that when you will pull request.


ebocher:
Close merged in the next one

PR #85, opened 2019-05-16
Osm processing chains

ebocher:
This PR implements as a set of subprocesses the OSM data extract and tranform operations.
This PR introduces several fixes in AbstractTablesInitialization and InputDataFormatting that seem never worked.
The main process is called prepareOSM. It saves the working data and the result files in a folder set by the user.
All the chain works but there is some fixes todo in InputDataFormatting process because all values are not well initialized. see vegetation type or building layer.
It's urgent to add some unit tests to ensure the quality of each processes.


pom.xml
```diff
@@ -7,7 +7,7 @@
     <parent>
         <groupId>org.orbisgis</groupId>
         <artifactId>orbisparent</artifactId>
-        <version>1.0.2-SNAPSHOT</version>
+        <version>1.0.3-SNAPSHOT</version>
```
SPalominos: Please, use instead v1.0.1. Versions 1.0.2 and higher uses H2GIS v2.0.0

ebocher:
Please wait

PR #86, opened 2019-05-16
Scales chain

ebocher:
about #84


j3r3m1:
Both tests prepareOSMDefault and PrepareOSMTest (used as processing chains) work but have an error message...
"the number of the input data map and the number of process input are different".

ebocher:
humm
I will check

PR #92, opened 2019-05-17
LCZ indicators are all calculated

j3r3m1:
The workflow is OK, the values seem OK.
Remains several things to modify:
- the tables join is currently manually performed (sql queries). It has to be changed when @ebocher will have coded the IProcess for joining tables
- some of the processes remove some RSU when they have no values (probably no building for example). It has to be modify in order to keep all RSU
- the LCZ identification algorithm should be developped. I look for that now


j3r3m1:
OK the merge with the previous PR works well (not conflicts). I am dealing with the debugging now since I use IProcess that I have modified in the previous PR.

PR #95, opened 2019-05-18
Replaced inner join by IProcess in LCZ processing chain

j3r3m1:


processing_chain/src/main/groovy/org/orbisgis/processingchain/BuildLCZ.groovy
```diff
@@ -234,22 +228,18 @@ public static createLCZ() {
...
+                         (calc_perviousness_frac.results.outputTableName)  : columnIdRsu,
+                         (calc_roughness_class.results.outputTableName)  : columnIdRsu]
+                                      , prefixName: prefixName+"_allindic", datasource: datasource])
 
                 datasource.execute("DROP TABLE IF EXISTS $rsu_indic0, $rsu_indic1, $rsu_indic2, $rsu_indic3".toString())
 
-                datasource.eachRow("SELECT * FROM ${rsu_all_indic}".toString()) {
+                datasource.eachRow("SELECT * FROM ${join_final_table.results.outputTableName}".toString()) {
                     row ->
                         println(row)
```
ebocher: Why a print here ?

processing_chain/src/main/groovy/org/orbisgis/processingchain/BuildLCZ.groovy
```diff
@@ -234,22 +228,18 @@ public static createLCZ() {
...
-                datasource.execute(("DROP TABLE IF EXISTS $rsu_all_indic; CREATE TABLE $rsu_all_indic AS SELECT a.*, " +
-                        "b.effective_terrain_roughness_class FROM $rsu_indic3 a INNER JOIN " +
-                        "${calc_roughness_class.results.outputTableName} b ON a.$columnIdRsu = b.$columnIdRsu").toString())
+                IProcess join_final_table = Geoclimate.DataUtils.joinTables()
+                join_final_table.execute([inputTableNamesWithId:
+                        [(join_for_SVF.results.outputTableName)       : columnIdRsu,
+                         (calc_aspect_ratio.results.outputTableName)  : columnIdRsu,
+                         (calcSVF.results.outputTableName)  : columnIdRsu,
+                         (calc_perviousness_frac.results.outputTableName)  : columnIdRsu,
+                         (calc_roughness_class.results.outputTableName)  : columnIdRsu]
```
ebocher: Write without parenthesis   calc_roughness_class.results.outputTableName:columnIdRsu

processing_chain/src/main/groovy/org/orbisgis/processingchain/BuildLCZ.groovy
j3r3m1: It was a temporary print since I wanted to check whether all indicators work well. The createLCZ is not finished yet, we need to implement the algorithm to identify the LCZ of each RSU. I propose I will remove the print when the algorithm will be implemented ?

processing_chain/src/main/groovy/org/orbisgis/processingchain/BuildLCZ.groovy
j3r3m1: I have initially done that but it does not work. It seems I can not write it without parenthesis when it is used as key in a map

PR #101, opened 2019-05-21
Geoindicators and fixes

ebocher:
This PR adds the block chain to compute indicators.
It introduces several fixes as validation for each processes on chains.
It offers also a new data util to save the tables as files.
The OSM extract process has been fixed to avoid null type on road and building.


ebocher:
humm don't understand why it doesn't compile here. Perhaps a time out on travis...
@SPalominos

SPalominos:
@ebocher It's caused by the 'executeScript' instruction which should use a Stream instead of a file path.

ebocher:
@SPalominos 
Dring

ebocher:
@SPalominos 
Feel free to merge this giant PR

ebocher:
Thanks let's start the last indicators

PR #104, opened 2019-05-23
Identify lcz

j3r3m1:
Still a problem to resolve with impervious fractions being to 0 for the MEDIAN method used for averaging.
Moreover, we need to separate more clearly the built LCZ from the natural ones (for example using the building_fraction).


j3r3m1:
I have an idea to deal with the problem of separation between built LCZ and natural LCZ but I think that first we should identify whether this mistake happens frequently on real datasets.

geoindicators/src/main/resources/org/orbisgis/LCZ_classes.csv
```diff
@@ -0,0 +1,18 @@
+name,sky_view_factor_inf,sky_view_factor_sup,aspect_ratio_inf,aspect_ratio_sup,building_surface_fraction_inf,building_surface_fraction_sup,impervious_surface_fraction_inf,impervious_surface_fraction_sup,pervious_surface_fraction_inf,pervious_surface_fraction_sup,height_of_roughness_elements_inf,height_of_roughness_elements_sup,terrain_roughness_class_inf,terrain_roughness_class_sup
```
ebocher: Is this file necessary ?

j3r3m1: not anymore we can remove it

ebocher:
Check failed

PR #119, opened 2019-06-07
BDTopo and Common iProcess works

gpetit:
Improve tests and other so that BDTopo and Common iProcess works
Thanks @j3r3m1

### DIFF
--- a/prepare_data/src/main/groovy/org/orbisgis/bdtopo/BDTopoGISLayers.groovy
+++ b/prepare_data/src/main/groovy/org/orbisgis/bdtopo/BDTopoGISLayers.groovy
@@ -64,18 +64,18 @@ static IProcess importPreprocess(){
                 logger.info('Executing the importPreprocess.sql script')
                 def uuid = UUID.randomUUID().toString().replaceAll('-', '_')
                 def tmpIris = 'TMP_IRIS_' + uuid
-                def zone = 'ZONE_' + uuid
+                def zone = 'ZONE'
                 def zoneBuffer = 'ZONE_BUFFER_' + uuid
                 def zoneExtended = 'ZONE_EXTENDED_' + uuid
-                def zoneNeighbors = 'ZONE_NEIGHBORS_' + uuid
+                def zoneNeighbors = 'ZONE_NEIGHBORS'
                 def bu_zone_indif = 'BU_ZONE_INDIF_' + uuid
                 def bu_zone_indus = 'BU_ZONE_INDUS_' + uuid
                 def bu_zone_remarq = 'BU_ZONE_REMARQ_' + uuid
-                def input_building = 'INPUT_BUILDING_' + uuid
-                def input_road = 'INPUT_ROAD_' + uuid
-                def input_rail = 'INPUT_RAIL_' + uuid
-                def input_hydro = 'INPUT_HYDRO_' + uuid
-                def input_veget = 'INPUT_VEGET_' + uuid
+                def input_building = 'INPUT_BUILDING'
+                def input_road = 'INPUT_ROAD'
+                def input_rail = 'INPUT_RAIL'
+                def input_hydro = 'INPUT_HYDRO'
+                def input_veget = 'INPUT_VEGET'
 
                 datasource.executeScript(getClass().getResourceAsStream('importPreprocess.sql'),
                         [ID_ZONE: idZone, DIST_BUFFER: distBuffer, EXPAND: expand, IRIS_GE: tableIrisName,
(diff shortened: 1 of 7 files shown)


prepare_data/src/main/groovy/org/orbisgis/bdtopo/BDTopoGISLayers.groovy
```diff
@@ -64,18 +64,18 @@ static IProcess importPreprocess(){
                 logger.info('Executing the importPreprocess.sql script')
                 def uuid = UUID.randomUUID().toString().replaceAll('-', '_')
                 def tmpIris = 'TMP_IRIS_' + uuid
-                def zone = 'ZONE_' + uuid
+                def zone = 'ZONE'
```
ebocher: Why delete the uuid value?

prepare_data/src/test/groovy/org/orbisgis/common/AbstractTablesInitializationTest.groovy
```diff
@@ -24,10 +23,11 @@ class AbstractTablesInitializationTest {
     @Test
     @EnabledIfSystemProperty(named = "test.bdtopo", matches = "true")
     void initParametersAbstract(){
-        H2GIS h2GISDatabase = H2GIS.open("./target/myh2gisbdtopodb")
+        H2GIS h2GISDatabase = H2GIS.open(bdTopoDb-".mv.db", "sa", "")
         def process = PrepareData.AbstractTablesInitialization.initParametersAbstract()
         assertTrue process.execute([datasource: h2GISDatabase])
         process.getResults().each {entry ->
+            println(entry)
```
ebocher: Delete println

prepare_data/src/test/groovy/org/orbisgis/common/InputDataFormattingTest.groovy
```diff
@@ -23,21 +23,38 @@ class InputDataFormattingTest {
...
+                         hLevMin: 3, hLevMax: 15, hThresholdLev2: 10, idZone: '56195',
 
                          buildingAbstractUseType: 'BUILDING_ABSTRACT_USE_TYPE', buildingAbstractParameters: 'BUILDING_ABSTRACT_PARAMETERS',
                          roadAbstractType: 'ROAD_ABSTRACT_TYPE', roadAbstractParameters: 'ROAD_ABSTRACT_PARAMETERS',
                          railAbstractType: 'RAIL_ABSTRACT_TYPE',
                          vegetAbstractType: 'VEGET_ABSTRACT_TYPE', vegetAbstractParameters: 'VEGET_ABSTRACT_PARAMETERS'])
         process.getResults().each {
-            entry -> assertNull h2GISDatabase.getTable(entry.getValue())
+            entry ->
+            println(entry)
```
ebocher: idem

prepare_data/src/test/groovy/org/orbisgis/common/InputDataFormattingTest.groovy
```diff
@@ -23,21 +23,38 @@ class InputDataFormattingTest {
     @Test
     @EnabledIfSystemProperty(named = "test.bdtopo", matches = "true")
     void inputDataFormatting(){
```
ebocher: The test is a good way but we must test also if there are some rows in the formatted tables

prepare_data/src/test/groovy/org/orbisgis/common/AbstractTablesInitializationTest.groovy
gpetit: Yes good catch. It's done in the following commit ;-)

prepare_data/src/test/groovy/org/orbisgis/common/InputDataFormattingTest.groovy
gpetit: Same. Removed ;-)

prepare_data/src/test/groovy/org/orbisgis/common/InputDataFormattingTest.groovy
gpetit: Yes, it's in the pipe. 
We are currently working to finalize all the cables and to check if tables are well created.
Then we will check if they are empty.

prepare_data/src/main/groovy/org/orbisgis/bdtopo/BDTopoGISLayers.groovy
j3r3m1: The uuid was sometimes used also for the output Tables while it is only useful for intermediate tables.

PR #122, opened 2019-06-07
Update some tests and improve the BD Topo processing chain

gpetit:
Update some tests and improve the BD Topo processing chain
@j3r3m1


gpetit:
Do not merge. It's just to upload the code so that @j3r3m1 will be able to continue

PR #123, opened 2019-06-07
Change the location of the DBTOPO database for test

SPalominos:
Change the location of the DBTOPO database for test which should be place in geoclimate/prepare_data/src/test/resources/org/orbisgis/common/myh2gisbdtopodb.mv.db.

### DIFF
--- a/prepare_data/src/test/groovy/org/orbisgis/bdtopo/BDTopoGISLayersTest.groovy
+++ b/prepare_data/src/test/groovy/org/orbisgis/bdtopo/BDTopoGISLayersTest.groovy
@@ -4,27 +4,42 @@ import org.junit.jupiter.api.BeforeAll
 import org.junit.jupiter.api.Test
 import org.junit.jupiter.api.condition.EnabledIfSystemProperty
 import org.orbisgis.PrepareData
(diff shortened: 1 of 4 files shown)


prepare_data/src/test/groovy/org/orbisgis/bdtopo/BDTopoGISLayersTest.groovy
```diff
@@ -4,27 +4,42 @@ import org.junit.jupiter.api.BeforeAll
...
+                os = new FileOutputStream(bdTopoDb)
+                byte[] buffer = new byte[1024]
+                int length
+                while ((length = is.read(buffer)) > 0) {
+                    os.write(buffer, 0, length)
+                }
+            } finally {
+                is.close()
+                os.close()
+            }
```
j3r3m1: I do not really understand what you are doing from line 21 up to the end. Could you explain or comment in the code ?

SPalominos: I have changed this piece of code in order to use groovy code and I've added comments.

Issue #114, opened 2019-05-28
Merging the geometries : long time


j3r3m1:
For information, it seems that the operation that does "Merging the geometries..." is also long (I have runned it for several territories and it seems even longer than the "Buildings preparation starts" that seems to last longer than 3 hours for Paris - not completed).
The associated IProcess is the SpatialUnits.createBlocks(). Maybe due to the buffer used, I do not know...

PR #126, opened 2019-06-16
Improve block processing

ebocher:
Now it uses graph technique to cluster the geometries.
It consumes less memory and it's faster than the previous method (100 X faster). 

About #114


geoindicators/src/main/groovy/org/orbisgis/SpatialUnits.groovy
```diff
@@ -202,22 +205,63 @@ static IProcess prepareRSUData(){
...
-                datasource.execute "CREATE TABLE $outputTableName as  select  EXPLOD_ID as $columnIdName, the_geom "+
-                        "from st_explode ('(select ST_UNION(ST_ACCUM(ST_BUFFER(THE_GEOM,$distance))) as the_geom"+
-                        " from $inputTableName)')"
-                logger.info("The geometries have been merged")
+
+
+                //Find all neighbors for each building
+                logger.info("Building index to perform the process...")
+                if(!datasource.getSpatialTable(inputTableName).the_geom.indexed){
+                    datasource.getSpatialTable(inputTableName).the_geom.createSpatialIndex()
```
SPalominos: Indexes are created only if they not already exists so you don't need to test that before.

geoindicators/src/main/groovy/org/orbisgis/SpatialUnits.groovy
```diff
@@ -202,22 +205,63 @@ static IProcess prepareRSUData(){
...
+                }
+                if(!datasource.getSpatialTable(inputTableName).id_build.indexed){
+                    datasource.getSpatialTable(inputTableName).id_build.createIndex()
```
SPalominos: Same thing

geoindicators/src/test/groovy/org/orbisgis/SpatialUnitsTests.groovy
```diff
@@ -71,7 +71,7 @@ class SpatialUnitsTests {
         assertTrue blockP.execute([inputTableName: "build_tempo",distance:0.01,
                      prefixName: "block", datasource: h2GIS])
         def outputTable = blockP.results.outputTableName
-        def countRows = h2GIS.firstRow "select count(*) as numberOfRows from $outputTable"
+        def countRows = h2GIS.firstRow "select count(*) as numberOfRows from $outputTable".toString()
```
SPalominos: `toString()` is not mandatory, you can remove it to simplify.

geoindicators/src/main/groovy/org/orbisgis/SpatialUnits.groovy
ebocher: I will fix it

geoindicators/src/test/groovy/org/orbisgis/SpatialUnitsTests.groovy
ebocher: Ok

PR #128, opened 2019-06-24
Fix srid for all the layers.

ELSW56:
Formerly, only rails had EPSG 2154.


ebocher:
Close because we will use the next OSM module

PR #133, opened 2019-07-18
geoclimate converted for being compatible with H2GIS2.0

j3r3m1:
Note that some of the code does not work due to
- the fact that OSM is not fixed
- the fact that ST_FORCE2D was issued with H2GIS


geoindicators/pom.xml
```diff
@@ -42,7 +42,7 @@
         <dependency>
             <groupId>org.orbisgis</groupId>
             <artifactId>h2gis-network</artifactId>
-            <version>1.5.0</version>
+            <version>2.0.0-20190705.044234-66</version>
```
ebocher: what ?

PR #147, opened 2019-09-19
New IProcess to calculate the free facade fraction within an extended…

j3r3m1:
… RSU (useful for a fast SVF calculation)


ebocher:
Seems there is an error on a unit test

PR #165, opened 2019-10-02
Add input parameters in some IProcess in processing chain

j3r3m1:
input parameters in some IProcess in processing chain


j3r3m1:
excellent your are brilliant

PR #166, opened 2019-10-02
Delete intermediate tables

j3r3m1:


geoindicators/src/main/groovy/org/orbisgis/geoindicators/Geoindicators.groovy
```diff
@@ -24,6 +24,15 @@ abstract class Geoindicators extends GroovyProcessFactory {
     //Utility methods
     static def getUuid(){
         UUID.randomUUID().toString().replaceAll("-", "_") }
+    static def getOutputTableName(prefixName, baseName){
+        if (prefixName == ""){
```
ebocher: if(prefixName)

PR #169, opened 2019-10-03
Svf last but not last

ebocher:
Do not merge


j3r3m1:
trouble with a extra input parameter added in spatial units (prepare RSU data)

PR #175, opened 2019-10-10
Add new tests on BD Topo

gpetit:
Add some tests (some were suggested by @j3r3m1 thx) mainly to check if objects are well selected or not in the input dataset

### DIFF
--- a/prepare_data/src/test/groovy/org/orbisgis/common/InputDataFormattingTest.groovy
+++ b/prepare_data/src/test/groovy/org/orbisgis/common/InputDataFormattingTest.groovy
@@ -221,6 +221,55 @@ class InputDataFormattingTest {
         assertEquals('building', h2GISDatabase.firstRow("SELECT TYPE FROM BUILDING " +
                 "WHERE ID_SOURCE='BATIMENT0000000290120079';")["TYPE"])
 
+        // Check if building are well selected or not ...
+        // ... with the building (INDIF) 'BATIMENT0000000290126798' which is inside the zone --> so expected 1
+        assertEquals(1, h2GISDatabase.firstRow("SELECT COUNT(*) as TOTAL FROM BUILDING " +
+                "WHERE ID_SOURCE='BATIMENT0000000290126798';")["TOTAL"])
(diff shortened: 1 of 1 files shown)


prepare_data/src/test/groovy/org/orbisgis/common/InputDataFormattingTest.groovy
```diff
@@ -762,6 +848,20 @@ class InputDataFormattingTest {
...
                 "WHERE ID_SOURCE='ZONEVEGE0000000222262077';")["HEIGHT_CLASS"])
 
+        // Check if vegetation area are well selected or not ...
+        // ... with the veget area 'ZONEVEGE0000000222259852' which is inside the zone --> so expected 1
+        assertEquals(1, h2GISDatabase.firstRow("SELECT COUNT(*) as TOTAL FROM VEGET " +
+                "WHERE ID_SOURCE='ZONEVEGE0000000222259852';")["TOTAL"])
+        // ... with the veget area 'ZONEVEGE0000000222246505' which is inside the extended zone --> so expected 1
+        assertEquals(1, h2GISDatabase.firstRow("SELECT COUNT(*) as TOTAL FROM VEGET " +
+                "WHERE ID_SOURCE='ZONEVEGE0000000222246505';")["TOTAL"])
+        // ... with the veget area 'ZONEVEGE0000000222257768' which is intersecting the extended zone (having a part outside) --> so expected 1
```
j3r3m1: for hydro and vegetation we merge geometries that are in contact and then could keep some of the resulting geometries that are splitted by the zone geometry depending on the vegetation (or hydro) total size. Is it possible to also test that ?

gpetit: We will indeed have to test this, but in my opinion it should not be done here, but rather when the RSUs are created.
Indeed, the cumulative area criterion (for objects that touch) is only used for RSUs.
Here we select the water and vegetation surfaces on a larger area, which has been made to select those objects that could be far from the municipality but still need to be taken into account.

j3r3m1: You are right here we keep it without any condition of size. Only in the RSU creation we merge and then test their size.

PR #189, opened 2019-10-18
Add more test on format OSM GIS layers

ebocher:
Set roof, wall and nb_level to 0 when one of them is negative
Do not compute  free_wall when nb_level=0
Fix negative width for road, replace to 0
Add auto_server=true for all tests


ebocher:
Do not merge it

Issue #185, opened 2019-10-17
Create table data type


ebocher:
Change double type in create table by double precision to be compliant with postgresql

ebocher:
duplicate with #195

Issue #183, opened 2019-10-16
Metadata files


ebocher:
Generate for each output files some metadata :  a readme or a pdf document that describe all variables stored in the tables and output files

gpetit:
Ok, I'm on it. So we will have two ways to document the outputs
- add comments in the tables (using the `COMMENT ON COLUMN schema.table.field IS 'xxxx';` syntax
- create a "user" doc, thanks to `.md` files that will be stored in the `/geoclimate/docs/` folder

gpetit:
I guess we can assume it's done, since the outputs are described in the online doc: https://github.com/orbisgis/geoclimate/blob/master/docs/chain_documentation/workflow/outputs.md

Issue #182, opened 2019-10-16
Styles


ebocher:
Add default SLD styles with the output results

- [x] Style for LCZ  ([see more](https://github.com/orbisgis/geoclimate/blob/master/docs/chain_documentation/classsifications/lcz.md#cartography)) 
- [x] Style for the Urban Typo ([see more](https://github.com/orbisgis/geoclimate/blob/master/docs/chain_documentation/classsifications/territorial_typologies.md#cartography)) 

The styles are stored here https://github.com/orbisgis/geoclimate/tree/master/processingchain/src/main/resources/styles

gpetit:
Not sure there is an "official" way to represent LCZs.
So what about the style we used here http://mapuce.orbisgis.org/ ?

gpetit:
They are the same. So the `lcz.sld` file is ready to be added in the project.
@ebocher in which folder do you want to store those kind of files ? Do we have like a `/styles/` folder ?

gpetit:
Ok. So the files will be stored in `/processing_chain/src/main/resources/styles/` folder.
We will have the following files (maybe some other will come after) :
- `lcz1.sld` : the style ready to be used for the field `LCZ1`in the `xxxx.RSU_LCZ` table
- `lcz2.sld` : the style ready to be used for the field `LCZ2`in the `xxxx.RSU_LCZ` table
- `lcz.png` : a graphic illustration of the expected result

The PR is in the pipe (with some other modifications)

j3r3m1:
brilliant

ebocher:
excelleeeeeeeeeeeeeent

PR #190, opened 2019-10-18
Add styles & doc and improve some sql scripts

gpetit:
- Add a `docs` folder and a first `README.md` file to start the user doc (see #183)
- Add `.sld` files for the LCZ (see #182)
- Add `precision` when using `double` fied (see #185)
- Filter input data coming from BD Topo in order to avoid building with a wall height < 0 and roads / rails with a ZINDEX < 0 (see #189)


docs/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+**This documentation is under construction.**
+
+
+
+## The Geoclimate chain
+
+
+
+Geoclimate chain is a collection of spatial processes to produce vector maps of cities
```
ebocher: Copy paste the main presentation available at the root of the repo

docs/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+
+
+## Inputs
+
+The Geoclimate chain is currently working with two kind of dataset : OpenStreetMap, on a worldwide context and BD Topo (from [IGN](http://ign.fr/)), only on a the french scale.
```
ebocher: Add bdtopo version

docs/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+
+
+### BD Topo (only in France)
+
+
+
+## Outputs
+
+As a result to the Geoclimate chain, 4 layers are produced :
```
ebocher: It depends on the chain used. 
There is a lot of missing information in this readme. I think we have to wait until it is more complete to validate it.

processing_chain/src/main/resources/styles/lcz1.sld
```diff
@@ -0,0 +1,286 @@
+<?xml version="1.0" encoding="UTF-8"?><sld:StyledLayerDescriptor xmlns="http://www.opengis.net/sld" xmlns:sld="http://www.opengis.net/sld" xmlns:gml="http://www.opengis.net/gml" xmlns:ogc="http://www.opengis.net/ogc" version="1.0.0">
+  <sld:NamedLayer>
+    <sld:Name>rsu_lcz</sld:Name>
+    <sld:UserStyle>
+      <sld:Name>rsu_lcz</sld:Name>
+      <sld:FeatureTypeStyle>
+        <sld:Name>name</sld:Name>
+        <sld:Rule>
+          <sld:Name>LCZ 1: Compact high-rise</sld:Name>
+          <MaxScaleDenominator>500000</MaxScaleDenominator>
```
ebocher: Do we need to keep the max scale here ?

processing_chain/src/main/resources/styles/lcz2.sld
```diff
@@ -0,0 +1,286 @@
+<?xml version="1.0" encoding="UTF-8"?><sld:StyledLayerDescriptor xmlns="http://www.opengis.net/sld" xmlns:sld="http://www.opengis.net/sld" xmlns:gml="http://www.opengis.net/gml" xmlns:ogc="http://www.opengis.net/ogc" version="1.0.0">
+  <sld:NamedLayer>
+    <sld:Name>rsu_lcz</sld:Name>
+    <sld:UserStyle>
+      <sld:Name>rsu_lcz</sld:Name>
+      <sld:FeatureTypeStyle>
+        <sld:Name>name</sld:Name>
+        <sld:Rule>
+          <sld:Name>LCZ 1: Compact high-rise</sld:Name>
+          <MaxScaleDenominator>500000</MaxScaleDenominator>
```
ebocher: idem

docs/README.md
gpetit: It was supposed to be a starting file, but no problem, I will add more informations

docs/README.md
gpetit: That's why I wrote at the beginning "This documentation is under construction." ;-).
Up to you. I can remove the file if you prefer

processing_chain/src/main/resources/styles/lcz1.sld
gpetit: Hum, IMO yes since it make no sense to display RSUs outside this scale range (geometries will be too small)

docs/README.md
gpetit: This file will be pushed in a separated PR. So I will remove it for this PR

docs/README.md
ebocher: thxs

docs/README.md
ebocher: Sounds good

processing_chain/src/main/resources/styles/lcz1.sld
ebocher: Note  that SLD style is to display a map in GIS app not in a server with a big area

gpetit: +1 good catch. I will remove the max scale filtering

gpetit: done

docs/README.md
gpetit: done

PR #194, opened 2019-10-21
Geoclim chains

j3r3m1:


processing_chain/src/test/groovy/org/orbisgis/processingchain/ChainProcessMainTest.groovy
```diff
@@ -1,27 +1,45 @@
...
 import static org.junit.jupiter.api.Assertions.assertTrue
 
 
 class ChainProcessMainTest {
 
     public static Logger logger = LoggerFactory.getLogger(ChainProcessMainTest.class)
 
+
+    @Test
+    void OSMGeoIndicators() {
```
ebocher: Is this test necessary? Because it only executes the process

j3r3m1: In my opinion, it is useful if we make some modifications in the code and we want to debug the geoclimateChain.

ebocher: Sure but this test is already checked in the script ProcessingChainOSMTest and ProcessingChainBDTopoTest. So I guess it's redundant here

j3r3m1: not exactly the same since it is new to include the LCZ calculation at the end of the geoindicators.

j3r3m1:
Currently, this PR leads to a new indicator name problem: when we make unweighted or weighted operations from lower scale (for example average building height within a RSU or building density within a RSU), we meet two problems (both are due to the fact that we have automatic indicator naming)
1. for building_area_density, it is now area_density since the indicator name is based on the original indicator name at building scale (and we have deleted the "building" key word)
2. for building_area_density, we must actually use fraction instead of density (according to a rule set previously which says that a ratio between two similar units, when the result is included within 0 and 1 should be called a fraction

j3r3m1:
>     1. for building_area_density, it is now area_density since the indicator name is based on the original indicator name at building scale (and we have deleted the "building" key word)
We may propose to the user (us mainly at this level of accuracy) to add the name of the "lower_scale_object" (e.g. building) in order to build the indicator names

j3r3m1:
>     2. for building_area_density, we must actually use fraction instead of density (according to a rule set previously which says that a ratio between two similar units, when the result is included within 0 and 1 should be called a fraction
For this one, I propose to make a modification of the indicator name only when the defined processing chains are called. If the user create its own chain or use a process solely, he will have to deal with an indicator called "density" instead of "fraction".

processing_chain/src/test/groovy/org/orbisgis/processingchain/ChainProcessMainTest.groovy
j3r3m1: No I have added this part of process within the geoClimateChain called "OSMGeoIndicators".

ebocher: My fault ok

j3r3m1:
I have made all modifications needed in the @gpetit .ods table.
DO NOT MERGE yet, I have to check if all indicator names are OK and add the road_fraction_overground

SPalominos:
@j3r3m1 here you are!

ebocher:
I think this pr must be merged before the big refactoring.

j3r3m1:
I agree (but needs review)

PR #204, opened 2019-10-24
Fix possible null when building layer is formated.

ebocher:


ebocher:
Note that the last PR introduces several bugs and the processing chain is not working anymore

PR #209, opened 2019-10-25
Modify indic

j3r3m1:
Concavity and Convexity have been modified
Perkins Skill score name has been modified (Unequity) and a new indicator has been added concerning the building direction (Unicity).


ebocher:
Seems that one process can not find the BUILDING_DIRECTION_UNIQUENESS value.

ebocher:
I decided to merge this PR because it compiles on my latopo and @j3r3m1's also.
It seems that H2 as a serious pb with MVMap. I m not able to isolate the bug in the full chain. Its occurrence is not consistent.

PR #212, opened 2019-11-06
Osm workflow module

ebocher:
Fixes bugs
Improve unit tests
Refactor the Geoclimate chain process and use now the name Workflow to qualify all the chain


ebocher:
Always the same bug due to H2 nio

PR #213, opened 2019-11-07
Add docs

gpetit:
Add a "first version" documentation.

M3 and OSM use case are still missing (it's in the todo list)

Have a look to https://github.com/gpetit/geoclimate/tree/add_docs/docs for the online "demo"


j3r3m1:
Huge work. Is there any opportunity to comment or propose modifications ? Three solutions:
- use GitHub but the acceptation of modifications will be just horrible
- use NextCloud : the advantage is that we can propose modification AND add comments. The shortcoming is that it means creating a file for each md file
- use a collaborative markdown engine such as the ones described here : https://techwiser.com/online-markdown-editor/ (such as https://hackmd.io/ looks like overleaf but no possibility to comment - however it seems possible to sync with GitHub while it is the contrary for https://draftin.com/).

ebocher:
I vote for github event if it will be horrible.

ebocher:
In fact you are right,  it's very horrible

gpetit:
https://hackmd.io/ seems to be good. I'll create an account and have a look

gpetit:
Hum not simple since most of the online editor only accept simple notes.
For the moment I did not found any editor that accept a project (with subfolders, ...)

j3r3m1:
Are you sure stackedit does not allow folder management ? Because when I have tried on their website (the test version), we can add some folders and deal with several files clicking on the top left folder.

ebocher:
Do not merge

gpetit:
Of course. Don't worry ;-)

PR #218, opened 2019-11-20
Fix bad input type

ebocher:


j3r3m1:
also done by gpetit in its PR

ebocher:
good

PR #223, opened 2019-11-29
Random forest

j3r3m1:
Add the first IProcess to create the randomForest from a training dataset in a ITable.
It is not yet the final version since for the moment there is not file to download on the internet (I have just a local version).


geoindicators/pom.xml
```diff
@@ -45,6 +45,18 @@
             <artifactId>h2gis-network</artifactId>
             <version>2.0.0-SNAPSHOT</version>
         </dependency>
+        <dependency>
+            <groupId>com.github.haifengl</groupId>
+            <artifactId>smile-core</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.github.haifengl</groupId>
+            <artifactId>smile-netlib</artifactId>
```
ebocher: Is this dep necessary ?

geoindicators/src/main/groovy/org/orbisgis/orbisprocess/geoclimate/geoindicators/TypologyClassification.groovy
```diff
@@ -240,3 +248,75 @@ IProcess identifyLczType() {
...
+
+            // Calculate the prediction using the same sample in order to identify what is the
+            // data rate that has been well classified
+            int[] prediction = Validation.test(model, df)
+            int[] truth = df.apply(varToModel).toIntArray()
+            double accuracy = Accuracy.of(truth, prediction)
+            logger.info "The percentage of the data that have been well classified is : ${accuracy*100}%"
+
+            if (save){
+                XStream xs = new XStream()
```
ebocher: Perhaps use jackson instead of XStream because it's a new dep.
@SPalominos

geoindicators/src/main/groovy/org/orbisgis/orbisprocess/geoclimate/geoindicators/DataUtils.groovy
```diff
@@ -34,14 +35,27 @@ IProcess joinTables() {
             inputTableNamesWithId.each { key, value ->
                 if (alias == "a") {
                     columnKey = "$alias.${value}"
-                    columns = datasource.getTable(key).columnNames.collect {
-                        alias + "." + it
```
ebocher: This add-on must be set in a separate PR.

geoindicators/pom.xml
j3r3m1: necessary I do not think so but it seems it optimizes the matrix manipulation:
see http://haifengl.github.io/faq.html

geoindicators/src/main/groovy/org/orbisgis/orbisprocess/geoclimate/geoindicators/DataUtils.groovy
j3r3m1: OK I will try to make a diff between this commit and 841781e

j3r3m1: OK I did it manually but it should be OK

geoindicators/src/main/groovy/org/orbisgis/orbisprocess/geoclimate/geoindicators/TypologyClassification.groovy
ebocher: We will look on with @SPalominos  after adding dataframe supports in orbisdata

SPalominos:
@ebocher , @j3r3m1 As a new module of orbisdata is comming with the `DataFrame` concept, maybe we can wait for it before merging this PR ?
The module will make easier the conversion of a `ITable` into a `DataFrame` and it also handle the geometries.

ebocher:
Yes sure

j3r3m1:
No hurry we also need to make a decision concerning the models and training data repository.

PR #225, opened 2019-12-02
Improve the process when no input shp files

gpetit:
Improve the process when no input shp files

- if no IRIS_GE (or empty table) then stop
- if no other input shapefiles then create empty tables


preparedata/src/main/groovy/org/orbisgis/orbisprocess/geoclimate/preparedata/bdtopo/BDTopoGISLayers.groovy
```diff
@@ -112,7 +112,37 @@ IProcess importPreprocess(){
...
+            // Check if the input files are present
+
+            // If the IRIS_GE table does not exist or is empty, then the process is stopped
+            if(!datasource.hasTable(tableIrisName) || datasource.getTable(tableIrisName).isEmpty()){
+                logger.error 'The process has been stopped since the table IRIS_GE does not exist or is empty'
+                return null}
+
+            // If the following tables does not exists, we create corresponding empty tables
+            if(!datasource.hasTable(tableBuildIndifName)){
+                datasource.execute("CREATE TABLE $tableBuildIndifName (THE_GEOM geometry(geometry, 2154), ID varchar, HAUTEUR integer);")
```
ebocher: Custom SRID ???
What about if the user wants to process out of 2154 ?

preparedata/src/main/groovy/org/orbisgis/orbisprocess/geoclimate/preparedata/bdtopo/BDTopoGISLayers.groovy
```diff
@@ -112,7 +112,37 @@ IProcess importPreprocess(){
...
+            }
+            if(!datasource.hasTable(tableBuildIndusName)){
+                datasource.execute("CREATE TABLE $tableBuildIndusName (THE_GEOM geometry(geometry, 2154), ID varchar, HAUTEUR integer, NATURE varchar);")
```
ebocher: typed geometry ? polygon ?

preparedata/src/test/groovy/org/orbisgis/orbisprocess/geoclimate/preparedata/bdtopo/BDTopoGISLayersTest.groovy
```diff
@@ -267,6 +267,404 @@ class BDTopoGISLayersTest {
...
+        assertNotNull(tableName)
+        assertEquals(3, tableName.columnCount)
+        assertEquals(0, tableName.rowCount)
+        // Check if the column types are correct
+        assertEquals('GEOMETRY', tableName.getColumnsType('THE_GEOM'))
+        assertEquals('VARCHAR', tableName.getColumnsType('ID'))
+        assertEquals('VARCHAR', tableName.getColumnsType('NATURE'))
+    }
+
+
```
ebocher: remove blank spaces

gpetit: My bad ;-) Done

preparedata/src/main/groovy/org/orbisgis/orbisprocess/geoclimate/preparedata/bdtopo/BDTopoGISLayers.groovy
gpetit: The idea was from you ;-)
Three options :
- we let as it is
- we remove the SRID
- the SRID is added as an input parameter

ebocher: From me to use one SRID ? my bad.
I think the SRID must be checked from the input table.

gpetit: Ok. And if the user has no .PRJ file or if the input table has no SRID ? We don't do anything ?

preparedata/src/main/groovy/org/orbisgis/orbisprocess/geoclimate/preparedata/bdtopo/BDTopoGISLayers.groovy
gpetit: Ok so I will replace geometry by polygon (and linestring for roads and railways)

preparedata/src/main/groovy/org/orbisgis/orbisprocess/geoclimate/preparedata/bdtopo/BDTopoGISLayers.groovy
gpetit: Hum, the problem is more complex.
How to check the SRID on a table that does not exist (since this is the problem here -> if there is no table, then we want to create a corresponding empty one).

So, we have these options:
- check the SRID from other tables in the same folder (or schema)
    - but what do we do if 
        - there is different SRID ? Take the first one ? Set a default one ?
        - there is no SRID ? --> Force with a default one ?
- define the SRID as an input parameter and force all the files (tables) with it
    - if there is existing other SRID, do we reproject ?
    - if there is files (tables) with no PRJ (SRID), forcing with a default SRID may induce errors (e.g if the input files is in WGS84 (with no PRJ - so we don't have this information) and we force with 2154, it's not correct)
- explain to users that in their input folder (or schema), all the files (tables) have to have a same SRID --> we use the first SRID found

gpetit: One other option, after a short discussion with @j3r3m1, could be :

The user has to provide same PRJ files.

Then we make the following test :
```
IF all input files (or tables) have a PRJ file (SRID)
  THEN (IF the PRJ (SRID) are the same 
                  THEN we use the SRID and continue the process
                  ELSE error message and stop the process
               )
  ELSE error message and stop the process
```

ebocher: User must provide a valid PRJ that contains an EPSG code.

ebocher: So
Read the first table - check SRID, table.srid.
If SRID =0 then process without SRID
else check if the SRID is the same for all tables else throw exception

gpetit: Ok I will work on it

preparedata/src/main/groovy/org/orbisgis/orbisprocess/geoclimate/preparedata/bdtopo/BDTopoGISLayers.groovy
```diff
@@ -112,6 +112,32 @@ IProcess importPreprocess(){
...
+            // -------------------------------------------------------------------------------
+            // Control the SRIDs from input tables
+
+            def list = [tableBuildIndifName, tableIrisName, tableBuildIndusName, tableBuildRemarqName,
+                        tableRoadName, tableRailName, tableHydroName, tableVegetName,
+                        tableImperviousSportName, tableImperviousBuildSurfName,
+                        tableImperviousRoadSurfName, tableImperviousActivSurfName]
+
+            // The SRID is stored and initialized to -1
+            def srid = -1
```
ebocher: default SRID is 0

preparedata/src/main/groovy/org/orbisgis/orbisprocess/geoclimate/preparedata/bdtopo/BDTopoGISLayers.groovy
gpetit: Done.
Check the SRID of tables. 
- If all the SRID are the same, then use the SRID and process
- If differences, then stop the process.

In the doc, we will explain that .PRJ files are expected and have to be the same

preparedata/src/main/groovy/org/orbisgis/orbisprocess/geoclimate/preparedata/bdtopo/BDTopoGISLayers.groovy
```diff
@@ -112,6 +112,32 @@ IProcess importPreprocess(){
...
+
+            // For each tables in the list, we check the SRID and compare to the srid variable. If different, the process is stopped
+            for(String name : list){
+                def table = datasource.getTable(name)
```
ebocher: Should be more efficient to query the geometry_columns view with the list of tables
sql.row(select srid from geometry_columns where F_TABLE_NAME in('listoftables'));

preparedata/src/main/groovy/org/orbisgis/orbisprocess/geoclimate/preparedata/bdtopo/BDTopoGISLayers.groovy
gpetit: We chose -1 to show that precisely this is a value that will be modified

preparedata/src/main/groovy/org/orbisgis/orbisprocess/geoclimate/preparedata/bdtopo/BDTopoGISLayers.groovy
SPalominos: To be more clear (@ebocher don't forget to use markdown syntax to embed code), the previous groovy instruction return an `List` of `GroovyResultSet`. once you have the list, you can request it to get all the SRIDs.

@ebocher with your instruction, we can't get the name of the problematic table. Will the usage of SQL request does a by gain of time as this code section will be done once ?

SPalominos:
Can be merged as the failing tests are the same as usual.

ebocher:
This PR fails but doesn't because the master compiles. Investigation...

gpetit:
Can we merge ? @ebocher

ebocher:
Nop because it doesn't compile yet.

ebocher:
I will restart the travis job

gpetit:
Waouh. It compiles !!

PR #226, opened 2019-12-02
Bdtopo workflow

ebocher:
This PR adds a workflow to process the BDTopo V2.

The process needs as input a folder  that contains all BDTopo shapefiles (V2 version).

Then each commune is processed and at the end of each process the results are stored in geojson file with the following convention  : $inseeCode_$tableName.geojson

To use it run

IProcess process = ProcessingChain.Workflow.BBTOPO_V2()
        assertTrue(process.execute(datasource: datasource,
                inputFolder: "./target/bdtopofolder/",
            outputFolder :"./target/geoclimate_chain/"))

### DIFF
--- a/processingchain/src/main/groovy/org/orbisgis/orbisprocess/geoclimate/processingchain/Workflow.groovy
+++ b/processingchain/src/main/groovy/org/orbisgis/orbisprocess/geoclimate/processingchain/Workflow.groovy
@@ -28,9 +28,11 @@ import org.orbisgis.datamanager.JdbcDataSource
  * investigate the sensibility of the chain to some input parameters
  * @param mapOfWeights Values that will be used to increase or decrease the weight of an indicator (which are the key
  * of the map) for the LCZ classification step (default : all values to 1)
+ * @param hLevMin Minimum building level height
+ * @param hLevMax Maximum building level height
+ * @param hThresholdLev2 Threshold on the building height, used to determine the number of levels
  * @param outputFolder The path of the folder to store the results as geojson files. By default, the files are stored
  * in the subfolder "results" in the inputFolder
- *
  *
  * @return the path of the folder that contains the geojson files
  *
@@ -53,9 +55,10 @@ def BBTOPO_V2() {
                 mapOfWeights : ["sky_view_factor" : 1, "aspect_ratio": 1, "building_surface_fraction": 1,
                                 "impervious_surface_fraction" : 1, "pervious_surface_fraction": 1,
                                 "height_of_roughness_elements": 1, "terrain_roughness_class": 1],
+                hLevMin : 3, hLevMax: 15, hThresholdLev2: 10,
                 outputFolder:""
         outputs outputFolder: String
-        run {datasource, inputFolder, distance,indicatorUse, svfSimplified, prefixName, mapOfWeights, outputFolder ->
+        run {datasource, inputFolder, distance,indicatorUse, svfSimplified, prefixName, mapOfWeights, hLevMin, hLevMax, hThresholdLev2,outputFolder ->
 
             if(inputFolder){
                 def outputDir = new File(outputFolder)
@@ -101,8 +104,8 @@ def BBTOPO_V2() {
                                                     tableHydroName             : 'SURFACE_EAU', tableVegetName: 'ZONE_VEGETATION',
                                                     tableImperviousSportName   : 'TERRAIN_SPORT', tableImperviousBuildSurfName: 'CONSTRUCTION_SURFACIQUE',
                                                     tableImperviousRoadSurfName: 'SURFACE_ROUTE', tableImperviousActivSurfName: 'SURFACE_ACTIVITE',
-                                                    distBuffer                 : 500, expand: 1000, idZone: code,
-                                                    hLevMin                    : 3, hLevMax: 15, hThresholdLev2: 10
+                                                    distBuffer                 : 500, expand: distance, idZone: code,
+                                                    hLevMin                    : hLevMin, hLevMax: hLevMax, hThresholdLev2: hThresholdLev2
                         ])){
 
                              String buildingTableName = prepareBDTopoData.getResults().outputBuilding
@@ -122,19 +125,20 @@ def BBTOPO_V2() {
                                      buildingTable       : buildingTableName,    roadTable       : roadTableName,
                                      railTable           : railTableName,        vegetationTable : vegetationTableName,
                                      hydrographicTable   : hydrographicTableName,indicatorUse    : indicatorUse,
-                                     svfSimplified       : svfSimplified,        prefixName      : prefixName,
+                                     svfSimplified       : svfSimplified,        prefixName      : "zone_$code",
                                      mapOfWeights        : mapOfWeights)) {
                                  error "Cannot build the geoindicators for the area $code"
                              }
                              else{
                                  def tablesToSave = geoIndicators.getResults().collect {
-                                     datasource.save(it.value,"${outputDir.getAbsolutePath()}${File.separator}${code}_${it.value}.geojson")
+                                     datasource.save(it.value,"${outputDir.getAbsolutePath()}${File.separator}${it.value}.geojson")
+                                     datasource.execute "DROP TABLE IF EXISTS ${it.value};"
                                      }
                                  info "${code} has been processed"
                              }
 
                          }
-                            info "Number of area processed $index on $nbAreas"
+                            info "Number of areas processed $index on $nbAreas"
                     }
 
                     }
@@ -149,8 +153,6 @@ def BBTOPO_V2() {
                 }
                 return [outputFolder : outputDir.getAbsolutePath()]
             }
-
-
         }
     })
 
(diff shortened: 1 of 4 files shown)


processingchain/src/main/groovy/org/orbisgis/orbisprocess/geoclimate/processingchain/Workflow.groovy
```diff
@@ -7,6 +7,155 @@ import org.orbisgis.datamanager.JdbcDataSource
...
+ * Urban Context: Computational Performance and Accuracy Analysis of Two Open and Free GIS Tools. Climate ,
+ * MDPI, 2018, Urban Overheating - Progress on Mitigation Science and Engineering Applications, 6 (3), pp.60.
+ * --> Stewart, Ian D., and Tim R. Oke. "Local climate zones for urban temperature studies." Bulletin of the American
+ * Meteorological Society 93, no. 12 (2012): 1879-1900.
+ *
+ */
+def BBTOPO_V2() {
+    create({
+        title "Create all geoindicators from OSM data"
+        inputs datasource: JdbcDataSource, inputFolder: String, distance: 0,indicatorUse: ["LCZ", "URBAN_TYPOLOGY", "TEB"],
```
j3r3m1: We must be clear with the distance parameter (used to merge geometries in the spatial units creation - mostly blocks from buildings).
In the lowest IProcess, distance = 0.01 as default value.
In the geoindicator processing chain, distance = 0.01
At the workflow scale:
- OSM: the distance parameter is not an input anymore
- BdTopo_v2: it is set to 0 by default
Then I propose to set a default value to 0.01 everywhere.

processingchain/src/main/groovy/org/orbisgis/orbisprocess/geoclimate/processingchain/Workflow.groovy
```diff
@@ -7,6 +7,155 @@ import org.orbisgis.datamanager.JdbcDataSource
...
+                        info "$nbAreas areas will be processed"
+                        inseeCodes.eachWithIndex { code, index->
+                         if(prepareBDTopoData.execute([datasource                 : datasource,
+                                                    tableIrisName              : 'IRIS_GE', tableBuildIndifName: 'BATI_INDIFFERENCIE',
+                                                    tableBuildIndusName        : 'BATI_INDUSTRIEL', tableBuildRemarqName: 'BATI_REMARQUABLE',
+                                                    tableRoadName              : 'ROUTE', tableRailName: 'TRONCON_VOIE_FERREE',
+                                                    tableHydroName             : 'SURFACE_EAU', tableVegetName: 'ZONE_VEGETATION',
+                                                    tableImperviousSportName   : 'TERRAIN_SPORT', tableImperviousBuildSurfName: 'CONSTRUCTION_SURFACIQUE',
+                                                    tableImperviousRoadSurfName: 'SURFACE_ROUTE', tableImperviousActivSurfName: 'SURFACE_ACTIVITE',
+                                                    distBuffer                 : 500, expand: 1000, idZone: code,
```
j3r3m1: concerning all parameters having default values, I wonder if we really need to keep them in the processing chain or workflow IProcess when they are not defined as input of these workflow or processing chains. If we modify the default value distBuffer in the prepareBDTopoData IProcess for example, we will have to modify the value of the parameter for all IProcesses that call prepareBDTopo if we want to keep the same default value everywhere.

processingchain/src/main/groovy/org/orbisgis/orbisprocess/geoclimate/processingchain/Workflow.groovy
```diff
@@ -7,6 +7,155 @@ import org.orbisgis.datamanager.JdbcDataSource
...
+                             String vegetationTableName = prepareBDTopoData.getResults().outputVeget
+
+                             String zoneTableName = prepareBDTopoData.getResults().outputZone
+
+                             IProcess geoIndicators = GeoIndicators()
+                             if (!geoIndicators.execute( datasource          : datasource,           zoneTable       : zoneTableName,
+                                     buildingTable       : buildingTableName,    roadTable       : roadTableName,
+                                     railTable           : railTableName,        vegetationTable : vegetationTableName,
+                                     hydrographicTable   : hydrographicTableName,indicatorUse    : indicatorUse,
+                                     svfSimplified       : svfSimplified,        prefixName      : prefixName,
```
j3r3m1: It seems that even for two different inseecode, you will have the same output table names. Thus you will remove almost all resulting tables from the datasource. Is this made on purpose ?
If not I would propose to add the InseeCode in the prefixName such as prefixName = PrefixName +"_" + codeInsee.

processingchain/src/main/groovy/org/orbisgis/orbisprocess/geoclimate/processingchain/Workflow.groovy
```diff
@@ -7,6 +7,155 @@ import org.orbisgis.datamanager.JdbcDataSource
...
+                                     mapOfWeights        : mapOfWeights)) {
+                                 error "Cannot build the geoindicators for the area $code"
+                             }
+                             else{
+                                 def tablesToSave = geoIndicators.getResults().collect {
```
j3r3m1: Could we add the opportunity not to save the results as in numerous IProcess (save = true by default) ? This would also be useful do add it to the OSM Workflow I think.

processingchain/src/main/groovy/org/orbisgis/orbisprocess/geoclimate/processingchain/Workflow.groovy
ebocher: Yes right. Call conf is needed

processingchain/src/main/groovy/org/orbisgis/orbisprocess/geoclimate/processingchain/Workflow.groovy
ebocher: Call conf needed

processingchain/src/main/groovy/org/orbisgis/orbisprocess/geoclimate/processingchain/Workflow.groovy
ebocher: I don't think so... To be discussed.

ebocher: Done the tables are saved

processingchain/src/main/groovy/org/orbisgis/orbisprocess/geoclimate/processingchain/Workflow.groovy
ebocher: Done with the prefixname =  zone_codeinsee_

processingchain/src/main/groovy/org/orbisgis/orbisprocess/geoclimate/processingchain/Workflow.groovy
ebocher: Done extract same parameters

processingchain/src/main/groovy/org/orbisgis/orbisprocess/geoclimate/processingchain/Workflow.groovy
ebocher: Distance is used for envelope

processingchain/src/main/groovy/org/orbisgis/orbisprocess/geoclimate/processingchain/Workflow.groovy
```diff
@@ -7,6 +7,157 @@ import org.orbisgis.datamanager.JdbcDataSource
...
+    create({
+        title "Create all geoindicators from OSM data"
+        inputs datasource: JdbcDataSource, inputFolder: String, distance: 0,indicatorUse: ["LCZ", "URBAN_TYPOLOGY", "TEB"],
+                svfSimplified:false, prefixName: "",
+                mapOfWeights : ["sky_view_factor" : 1, "aspect_ratio": 1, "building_surface_fraction": 1,
+                                "impervious_surface_fraction" : 1, "pervious_surface_fraction": 1,
+                                "height_of_roughness_elements": 1, "terrain_roughness_class": 1],
+                hLevMin : 3, hLevMax: 15, hThresholdLev2: 10,
+                outputFolder:""
+        outputs outputFolder: String
```
j3r3m1: do we add a parameter 'save' in order to make the file saved or the ITable created depending on if it is true or false ?

processingchain/src/test/groovy/org/orbisgis/orbisprocess/geoclimate/processingchain/ProcessingChainBDTopoTest.groovy
```diff
@@ -243,4 +243,17 @@ class ProcessingChainBDTopoTest extends ChainProcessMainTest{
                 abstractTables.outputRoad, abstractTables.outputRail, abstractTables.outputVeget,
                 abstractTables.outputHydro, saveResults, svfSimplified, indicatorUse,  prefixName)
     }
+
+    //@Test
+    void testBDTOPO_V2Workflow() {
```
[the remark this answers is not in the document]
ebocher: This test is disable. It's use for debug purpose.

processingchain/src/main/groovy/org/orbisgis/orbisprocess/geoclimate/processingchain/Workflow.groovy
ebocher: Nop because you will have multiple tables in the database so I think it's better to save the results in file

j3r3m1: ok

PR #227, opened 2019-12-04
Updates OrbisData.

SPalominos:
Updates OrbisData.


ebocher:
There is a fail here https://travis-ci.com/orbisgis/geoclimate/jobs/262937229#L4261
due to the update

PR #236, opened 2019-12-06
Modify BDTopo_V2 default value

j3r3m1:
set distance to 1000 instead of 0

### DIFF
--- a/processingchain/src/main/groovy/org/orbisgis/orbisprocess/geoclimate/processingchain/Workflow.groovy
+++ b/processingchain/src/main/groovy/org/orbisgis/orbisprocess/geoclimate/processingchain/Workflow.groovy
@@ -51,7 +51,7 @@ import org.orbisgis.orbisdata.datamanager.jdbc.JdbcDataSource
 def BDTOPO_V2() {
     create({
         title "Create all geoindicators from BDTopo data"
-        inputs datasource: JdbcDataSource, inputFolder: String, distance: 0,indicatorUse: ["LCZ", "URBAN_TYPOLOGY", "TEB"],
+        inputs datasource: JdbcDataSource, inputFolder: String, distance: 1000,indicatorUse: ["LCZ", "URBAN_TYPOLOGY", "TEB"],
                 svfSimplified:false, prefixName: "",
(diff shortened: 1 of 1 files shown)


ebocher:
OK must be fixed in the futur because BDTOPO process should work without distance value.
A good way will be to separate data preparation and statistics (QA)
@gpetit

processingchain/src/main/groovy/org/orbisgis/orbisprocess/geoclimate/processingchain/Workflow.groovy
```diff
@@ -51,7 +51,7 @@ import org.orbisgis.orbisdata.datamanager.jdbc.JdbcDataSource
 def BDTOPO_V2() {
     create({
         title "Create all geoindicators from BDTopo data"
-        inputs datasource: JdbcDataSource, inputFolder: String, distance: 0,indicatorUse: ["LCZ", "URBAN_TYPOLOGY", "TEB"],
+        inputs datasource: JdbcDataSource, inputFolder: String, distance: 1000,indicatorUse: ["LCZ", "URBAN_TYPOLOGY", "TEB"],
```
ebocher: Not that this distance input is not used in inputdataformatting.sql because the value is set in the script see
ST_DIFFERENCE(ST_EXPAND(the_geom, 1000), the_geom)
@gpetit

gpetit: Note sure to understand.
The `EXPAND` parameters (default value = 1000 m) is used in
- `importPreprocess.sql` (so in `BDTopoGISLayers.groovy`)
- `inputDataFormatting.sql` (so in `inputDataFormatting.groovy`)

ebocher: I mean :
 ST_DIFFERENCE(ST_EXPAND(the_geom, $distance), the_geom)
distance must be a parameter for the inputdataformatting.sql

gpetit: Oups my bad, I forgot this one. I will make the modification

Issue #239, opened 2019-12-10
Update the syntax for SPATIAL INDEX


gpetit:
As seen in https://github.com/h2database/h2database/issues/2171, update the syntax for SPATIAL INDEX in all Geoclimate files and also the function in OrbisData

--> assigned to @gpetit 
@ebocher @SPalominos @j3r3m1

gpetit:
Concerning OrbisData, the fix has been made by @SPalominos (see https://github.com/orbisgis/orbisdata/pull/190)

PR #240, opened 2019-12-11
Fix java version.

SPalominos:
Fix java version.


SPalominos:
It would be added by the resolution of the issue #239

PR #257, opened 2019-12-17
Change the place where MakeValid is made

gpetit:
As requested in #252, change the place in the code where the `ST_MakeValid` is made. 

Now we operate this function directly at the beginning when using input layers coming from BD Topo.


ebocher:
Sure a test case must be added

gpetit:
There is already a test on vegetation, but I can add a test on building (I have to find a case)

j3r3m1:
the test case could be the territory use in issue #252 ?

ebocher:
Because ST_MakeValid transform the input data, I think that the statistics must be disable.

gpetit:
Ok I have a case on building for Vannes (the city used for tests).

BATI_INDIFFERENCIE
- ID : `BATIMENT0000000290122667`
- THE_GEOM : `MULTIPOLYGON (((268415.5 6744987.2, 268417.6 6744988, 268418.7 6744982.7, 268416.3 6744981.9, 268415.5 6744987.2), (268407.9 6744976, 268416.5 6744979.3, 268421.5 6744981.2, 268417.7 6744993.1, 268415.2 6744992.3, 268406.3 6744989.5, 268407.9 6744976)))`

MakeValid returns

`MULTIPOLYGON (((268407.9 6744976, 268406.3 6744989.5, 268415.2 6744992.3, 268417.7 6744993.1, 268421.5 6744981.2, 268416.5 6744979.3, 268407.9 6744976), (268415.5 6744987.2, 268416.3 6744981.9, 268418.7 6744982.7, 268417.6 6744988, 268415.5 6744987.2)))`

So I will add a test on this building

j3r3m1:
do you add a test only on this specific building or directly test the importpreprocess on the city of Vannes ?

gpetit:
Not sure to understand. I will add a test on this specific building, but in the BDTopoGISLayersTest.groovy file in which we have all the test regarding the preparation of BD Topo, before going in the inputDataFormating part.

I just want to make sure that this input building from IGN, which is not valid, becomes valid in the `INPUT_BUILDING` table

j3r3m1:
So you will apply the process on all the city of Vannes and then verify that this specific building is valid ?

gpetit:
I just added the test. 
It was simplest than expected. I just check that the building (coming from `BATI_INDIFFERENCIE` and known to be invalid) is now valid (in the `INPUT_BUILDING` table)

j3r3m1:
So simple that now the checks fail. =)
Strange error however it seems that it comes from the OSM side. I have re-run Travis, hopefully it was only a bug due to Travis... Let's see

j3r3m1:
OK, even if the tests are currently only on Vannes and not on every city where we got problems, I validate it since it passes and because we will change the settings to make possible the testing on several cities. I do that in the begining of the afternoon and @gpetit will be able to add specific tests for specific territories.

Issue #254, opened 2019-12-16
Workflow BD Topo - 84007


j3r3m1:
The chain seems to run correctly for all indicators but then I got the following non verbose error:
> [INFO] org.orbisgis.orbisprocess.geoclimate.processingchain.ProcessingChain - Geoindicators calculation time: 63.447 s
[INFO] org.orbisgis.orbisprocess.geoclimate.processingchain.ProcessingChain - All geoindicators have been computed
[ERROR] org.orbisgis.orbisdata.processmanager.process.Process - Error while executing the process.
null
 
The error happens in the Geoindicator Workflow after this line:
https://github.com/orbisgis/geoclimate/blob/master/processingchain/src/main/groovy/org/orbisgis/orbisprocess/geoclimate/processingchain/Workflow.groovy#L355

It seems that the error is process related, may be when trying to execute the last process of the chain (the LCZ):
https://github.com/orbisgis/geoclimate/blob/master/processingchain/src/main/groovy/org/orbisgis/orbisprocess/geoclimate/processingchain/Workflow.groovy#L381

@SPalominos an idea where does this non verbose error could come from ?

j3r3m1:
OK I think I got it. It can not come from the LCZ process since it is not used in my chain (I select only URBAN_TYPOLOGY).
Thus the rsuLCZ (output from geocindicator workflow) is set to null, which becomes a problem when the BD Topo workflow tries to save all output from geoindicators...
We must add a null value condition at this row (when the output are saved...)
https://github.com/orbisgis/geoclimate/blob/master/processingchain/src/main/groovy/org/orbisgis/orbisprocess/geoclimate/processingchain/Workflow.groovy#L134

j3r3m1:
done according to #258

PR #258, opened 2019-12-17
save only tables that are not null (solve issue #254

j3r3m1:


ebocher:
Is there a test case for this fix ;-)

j3r3m1:
=) I thought @gpetit would have ask that before you. Then should I uncomment the tests that are already existing in the geoclimate project and use them for this specific case ?

ebocher:
ok

j3r3m1:
OK I could not make the current workflow test work because of the inputFolder parameter which is not recognized as a directory. I have got not idea how to make it work

processingchain/src/main/groovy/org/orbisgis/orbisprocess/geoclimate/processingchain/Workflow.groovy
```diff
@@ -131,7 +131,7 @@ def BDTOPO_V2() {
                              }
                              else{
                                  def tablesToSave = geoIndicators.getResults().collect {
-                                     if (datasource.hasTable(it)){
+                                     if (it.value != null){
```
SPalominos: You can replace `if(it.value != null)` by `if(it.value)` (for more info : [Groovy truth](https://groovy-lang.org/semantics.html#Groovy-Truth))

processingchain/src/test/groovy/org/orbisgis/orbisprocess/geoclimate/processingchain/ProcessingChainBDTopoTest.groovy
```diff
@@ -13,25 +13,19 @@ import static org.junit.jupiter.api.Assertions.*
 class ProcessingChainBDTopoTest extends ChainProcessAbstractTest{
 
     public static Logger logger = LoggerFactory.getLogger(ProcessingChainBDTopoTest.class)
+    public static String h2db = "./target/myh2gisbdtopodb;AUTO_SERVER=TRUE"
+    public static String bdtopoFoldName = "bdtopofolder"
+    public static String[] listTables = ["IRIS_GE", "BATI_INDIFFERENCIE", "BATI_INDUSTRIEL", "BATI_REMARQUABLE",
```
SPalominos: If you don't need to explicitly type a variable, use `def` instead of class.

processingchain/src/test/groovy/org/orbisgis/orbisprocess/geoclimate/processingchain/ProcessingChainBDTopoTest.groovy
```diff
@@ -58,10 +52,39 @@ class ProcessingChainBDTopoTest extends ChainProcessAbstractTest{
         }
     }
 
+    def loadFiles(String inseeCode){
+        H2GIS h2GISDatabase = H2GIS.open(h2db, "sa", "")
```
SPalominos: Be careful, by using a fixed String as database name, you will always use the same database for each tests. If needed, to avoid conflict, you might have to use random names (by using `UUID` class).

processingchain/src/test/groovy/org/orbisgis/orbisprocess/geoclimate/processingchain/ProcessingChainBDTopoTest.groovy
```diff
@@ -58,10 +52,39 @@ class ProcessingChainBDTopoTest extends ChainProcessAbstractTest{
...
+        String relativePath = bdtopoFoldName + File.separator + inseeCode
+
+        // Test whether there is a folder containing .shp files for the corresponding INSEE code
+        if(ProcessingChain.class.getResource(relativePath) != null){
+            // Test is the URL is a folder
+            if(new File(ProcessingChain.class.getResource(relativePath).toURI()).isDirectory()){
+                listTables.each {
+                    String filePath = ProcessingChain.class.getResource(relativePath + File.separator + it + ".shp")
```
SPalominos: It might fail because `getResource()` return an `URL` and you are forcing Groovy to cast it as `String`. You should use `def filePath = ...`.

processingchain/src/test/groovy/org/orbisgis/orbisprocess/geoclimate/processingchain/ProcessingChainBDTopoTest.groovy
```diff
@@ -58,10 +52,39 @@ class ProcessingChainBDTopoTest extends ChainProcessAbstractTest{
...
+                    // If some layers are missing, do not try to load them...
+                    if (filePath != null) {
+                        h2GISDatabase.load(filePath, it, true)
+                    }
+                }
+            }
+            else{
+                logger.error  "There is no folder containing shapefiles for commune insee $inseeCode"
```
SPalominos: If in this case, the database is not well generated and doesn't contains the needed data, you can do a `fail()` (from `org.junit.jupiter.api`) to indicate to JUnit that there is an error and stop the test here.

processingchain/src/test/groovy/org/orbisgis/orbisprocess/geoclimate/processingchain/ProcessingChainBDTopoTest.groovy
```diff
@@ -58,10 +52,39 @@ class ProcessingChainBDTopoTest extends ChainProcessAbstractTest{
...
+            }
+        }
+        else{
+            logger.error  "There is no folder containing shapefiles for commune insee $inseeCode"
```
SPalominos: Same as above

processingchain/src/test/groovy/org/orbisgis/orbisprocess/geoclimate/processingchain/ProcessingChainBDTopoTest.groovy
```diff
@@ -240,16 +263,27 @@ class ProcessingChainBDTopoTest extends ChainProcessAbstractTest{
...
-        IProcess process = ProcessingChain.Workflow.BDTOPO_V2()
-        assertTrue(process.execute(datasource: datasource,
-                inputFolder: "./target/bdtopofolder/",
-                outputFolder :"./target/geoclimate_chain/"))
+        IProcess processBDTopo = ProcessingChain.Workflow.BDTOPO_V2()
+        assertTrue(processBDTopo.execute(datasource: datasource,
+                inputFolder: inpFolder.getAbsolutePath().toString(),
+                outputFolder :"./target/geoclimate_chain/",
+                indicatorUse: ["TEB"]))
+        String baseNamePathAndFileOut = processBDTopo.getResults().outputFolder + File.separator + "zone_" + inseeCode + "_"
```
SPalominos: You should assert that `processBDTopo.getResults().outputFolder` is not null

processingchain/src/main/groovy/org/orbisgis/orbisprocess/geoclimate/processingchain/Workflow.groovy
j3r3m1: done

processingchain/src/test/groovy/org/orbisgis/orbisprocess/geoclimate/processingchain/ProcessingChainBDTopoTest.groovy
j3r3m1: done

processingchain/src/test/groovy/org/orbisgis/orbisprocess/geoclimate/processingchain/ProcessingChainBDTopoTest.groovy
j3r3m1: OK I have corrected that using specific name for each test (to be able to easily find the right db when applying a test...)

processingchain/src/test/groovy/org/orbisgis/orbisprocess/geoclimate/processingchain/ProcessingChainBDTopoTest.groovy
j3r3m1: done

processingchain/src/test/groovy/org/orbisgis/orbisprocess/geoclimate/processingchain/ProcessingChainBDTopoTest.groovy
j3r3m1: along with the logger or only use the fail ?

processingchain/src/test/groovy/org/orbisgis/orbisprocess/geoclimate/processingchain/ProcessingChainBDTopoTest.groovy
j3r3m1: just assert this or add a condition to the initialisation of the variable and to the following assertions ?

SPalominos: Just an assertion

processingchain/src/test/groovy/org/orbisgis/orbisprocess/geoclimate/processingchain/ProcessingChainBDTopoTest.groovy
SPalominos: Along with the logger.

SPalominos:
@j3r3m1 :tada: Well done, you PR made the Jenkins build ~~great~~ work again

PR #274, opened 2020-02-10
Osm configuration file

ebocher:
As for BDTopo now we have a configuration file to custom extraction and transformation


ebocher:
I m on it

PR #286, opened 2020-02-21
Make case delete_db = false work

j3r3m1:


processingchain/src/main/groovy/org/orbisgis/orbisprocess/geoclimate/processingchain/Workflow.groovy
```diff
@@ -1568,7 +1568,7 @@ def OSM() {
                 if(geoclimatedb){
                     def h2gis_path = geoclimatedb.get("path")
                     def delete_h2gis_db = geoclimatedb.get("delete")
-                    if(delete_h2gis_db && delete_h2gis_db in Boolean){
+                    if((delete_h2gis_db && delete_h2gis_db) in Boolean){
```
ebocher: humm. I will look for rigth now

ebocher: Seems better to do

 if((delete_h2gis_db && (delete_h2gis_db.equalsIgnoreCase("false") || delete_h2gis_db.equalsIgnoreCase("true")){

}

j3r3m1: I do not understand, I got error message using equalsIgnoreCase("false"): it is supposed to work only with string, not boolean.

PR #288, opened 2020-02-22
Add tests for verifying the indicatorUse values are not the default values

j3r3m1:
I did not add real testing on the columns of the resulting database, I have just verified that almost no indicator were used and it seems OK. Not sure it is necessary to merge actually


ebocher:
It will be better to create the config file from the code so you can custom output folder and database path. See an example here https://github.com/orbisgis/geoclimate/blob/8d946deb665fc08f732db1d98c54803523ad5a95/processingchain/src/test/groovy/org/orbisgis/orbisprocess/geoclimate/processingchain/ProcessingChainOSMTest.groovy#L223

ebocher:
I'm on it

PR #296, opened 2020-02-26
Fix trouble and add test

### LINKED ISSUE #292: Nb_Building not described
Several indicators described in docs/chain_documentation/indicators/rsu.md are involving another indicator named NB_Building but this one is not defined.

j3r3m1:
Fix #292 
Note that a default value of "id_build" has been set for the inputLowerScale. It was to avoid modifying all processes using this IProcess but I can change that if needed (will take a bit more time).
I will further investigate other functions using count since they may be also wrong.


ebocher:
Do we need to wait a new fix to not use id_build as default column name ?

j3r3m1:
nop, it is supposed to be done. There is no default column name anymore

j3r3m1:
OK sorry it did not push due to a problem of sync between the origin and my local version

Issue #302, opened 2020-02-28
RSU_SMALLEST_UNITS duplicates


j3r3m1:
When applying the RSU_SMALLEST_UNITS, it does not work as expected (at least concerning the vegetation layer). For several RSU, I get a "low_vegetation density" (but similarly for the high vegetation) higher than 1.
When I investigate the input layer, I found out that I may have several low_vegetation geometries covering each other. I understand that they may have a different type and then a different use. However when applying the rsu_smallest_units, for now we only want to know whether a place is covered by low or high (or both) vegetation. Then I suppose we would need to union low vegetation that intersect each other (and same for high vegetation).
I suppose that to work, the all "low vegetation" objects should be union when intersect

j3r3m1:
I have tried a st_union(st_accum on vegetation geometries but not sufficient (sql ask me to put the a.the_geom in the group by list...
```sql
CREATE TABLE $high_vegetation_tmp AS SELECT st_intersection(ST_FORCE2d(a.the_geom), st_union(st_accum(b.the_geom))) AS the_geom, b.id_rsu FROM 
                            $vegetationTable AS a, $rsuTable AS b WHERE a.the_geom && b.the_geom 
                            AND ST_INTERSECTS(a.the_geom, b.the_geom) AND a.height_class='high' GROUP BY b.id_rsu
```

ebocher:
I'm on it

j3r3m1:
Ok work

j3r3m1:
Thanks to  #305

Issue #301, opened 2020-02-28
RSU_SMALLEST_UNITS height class


j3r3m1:
I apply the rsu_smallest_unit process only with vegetation and RSU and I do not get any information in the output table.
I have debugged the function and I came out with the following problem: 
https://github.com/orbisgis/geoclimate/blob/24bc45a5d2dc0f5b1b6ef185da14da9bc88df82e/geoindicators/src/main/groovy/org/orbisgis/orbisprocess/geoclimate/geoindicators/RsuIndicators.groovy#L1481
Should be height_class instead of type I suppose ?

ebocher:
Goodddddddddddddd catch.
I will fix it

j3r3m1:
OK work

PR #305, opened 2020-03-06
Fix RSU_SMALLEST_UNITS

ebocher:
About #301 and #302


j3r3m1:
OK still a problem with the fact that the intersections creates points and linestring then it is necessary to use a collection extract (for dimension 3) to make it work

j3r3m1:
do you want I make the correction ?

ebocher:
Do you have a bug ?

PR #307, opened 2020-03-06
Add collection extract 3

j3r3m1:
Correct SmallestCommunGeometry to extract only polygons (after RSU intersection)


ebocher:
You are so brillant

Issue #314, opened 2020-03-10
Aspect ratio crash when area_density = 1


gpetit:
I was processing the city 44083 with BD Topo v2 and it crashes with this error logs

```
...

!ENTRY groovy.lang.GroovyShell 1 0 2020-03-09 16:55:01.895
!MESSAGE Executing RSU aspect ratio

!ENTRY groovy.lang.GroovyShell 1 0 2020-03-09 16:55:01.902
!MESSAGE Cannot compute the aspect ratio calculation 

!ENTRY groovy.lang.GroovyShell 4 0 2020-03-09 16:55:01.903
!MESSAGE Cannot compute the RSU indicators

!ENTRY groovy.lang.GroovyShell 4 0 2020-03-09 16:55:01.904
!MESSAGE Cannot build the geoindicators for the area with the INSEE id_zone : 44083

!ENTRY groovy.lang.GroovyShell 1 0 2020-03-09 16:55:01.904
!MESSAGE 44083 has been processed
```

So we investigated with @SPalominos and we found that the problem seems to come from [this line](https://github.com/orbisgis/geoclimate/blob/f48595f00eaa7cf45a85735cdd68deb97fe16f5e/geoindicators/src/main/groovy/org/orbisgis/orbisprocess/geoclimate/geoindicators/RsuIndicators.groovy#L241) when the `area_density` (`$rsuBuildingDensityColumn`) = 1

We made a test in debug mode

```
datasource.execute """SELECT $rsuFreeExternalFacadeDensityColumn/(1-$rsuBuildingDensityColumn) AS $BASE_NAME, $COLUMN_ID_RSU FROM $rsuTable"""
```
Returns `Method threw 'org.codehaus.groovy.runtime.InvokerInvocationException' exception.`

Now the same, but filtering the values equal to 1

```
datasource.execute """SELECT $rsuFreeExternalFacadeDensityColumn/(1-$rsuBuildingDensityColumn) AS $BASE_NAME, $COLUMN_ID_RSU FROM $rsuTable where $rsuBuildingDensityColumn<>1"""
```
Returns `true`

j3r3m1:
What a bad luck, building density = 1. Then the aspect ratio means nothing.
It will probably be different using the new algorithm for surface fractions.
However, to avoid this bug again, I propose to set the aspect ratio to null when the building density is 1.

gpetit:
> I propose to set the aspect ratio to null when the building density is 1.

[...]

PR #316, opened 2020-03-11
Set aspect ratio to null when build density = 1

j3r3m1:
Modify process + add test
Link with #314


gpetit:
Wahou ! One reviewer say "Yes" ! Who will merge now ?! ;-)

j3r3m1:
Je vais me gêner tiens !

PR #321, opened 2020-03-27
Surface fraction in workflow

j3r3m1:
I have replaced the surface fraction independent IProcesses by a unique one in the workflow.

The indicators produced are different than the previous ones, thus needed to modify the workflow tests.

I have defined compulsory surface fractions indicators needed for the classifications (urban typo and lcz). In the workflow, the user has to define how he wants to build these indicators according to superposed surface fractions


j3r3m1:
@SPalominos any idea of the reason of the failure ? It is not a test error anymore, I have no clue to resolve that.

geoindicators/src/main/groovy/org/orbisgis/orbisprocess/geoclimate/geoindicators/TypologyClassification.groovy
```diff
@@ -53,12 +55,13 @@ IProcess identifyLczType() {
     
     return create({
         title "Set the LCZ type of each RSU"
-        inputs rsuLczIndicators: String, prefixName: String, datasource: JdbcDataSource, normalisationType: "AVG",
+        inputs rsuLczIndicators: String     , rsuOtherIndicators: String        ,prefixName: String,
+                datasource: JdbcDataSource  , normalisationType: "AVG"          ,
                 mapOfWeights: ["sky_view_factor"             : 1, "aspect_ratio": 1, "building_surface_fraction": 1,
                                "impervious_surface_fraction" : 1, "pervious_surface_fraction": 1,
                                "height_of_roughness_elements": 1, "terrain_roughness_class": 1]
         outputs outputTableName: String
```
ebocher: rsuOtherIndicators - > rsuAllIndicators

j3r3m1:
I have deleted the buildLCZ file (and then processing chain) since it is redundant with the geoindicator IProcess when set to indicator2Use = ["LCZ"]. Then we have to make a modification to both IProcess if one is modified.
I also modify a test that was not disabled and had a lot of typing errors (it seems). However, at the end, I realize that this test was to be performed on local machine since the config file to run for the workflow had no real path...

j3r3m1:
OK still have a problem with the oracle jdk version which does not want to pass the tests. @SPalominos any idea why ?

PR #331, opened 2020-04-16
New version of equality and uniqueness indic (sql free)

j3r3m1:
Try to improve the performance of the IProcess...


j3r3m1:
@gpetit  can you get this one and try if it does improve the workflow for bdtopo

j3r3m1:
OK tested on Nantes it is less than 30 minutes and the uniqueness and equality is almost "instantaneous". @gpetit for information I have modified the inequality indicator to "equality" indicator since closer the value is from 1, more equal it is.

gpetit:
my test is currently in progress

gpetit:
Done in 38mn (but my machine is older than yours)
And the resulting LCZ looks better (not perfect, but on the good way)

image

j3r3m1:
OK good to see that the "rural types" are much better performed than before. Let's work on the urban part then.

Issue #332, opened 2020-04-17
Update doc


ebocher:
Due the last modules refactoring https://github.com/orbisgis/geoclimate/pull/330, the doc must be updated

gpetit:
Ok, so if I understand well, to launch the chain,  we used to have

```groovy
@GrabResolver(name='orbisgis', root='https://nexus.orbisgis.org/repository/orbisgis/')
@Grab(group='org.orbisgis.orbisprocess', module='geoclimate', version='1.0.0-RC1')

import org.orbisgis.orbisprocess.geoclimate.Geoclimate
Geoclimate.logger = logger

def process = Geoclimate.Workflow.BDTOPO_V2()
process.execute(configurationFile:'/home/geoclimate/bdtopo/config_all_zones.json')
```

and now, we have
```groovy
@GrabResolver(name='orbisgis', root='https://nexus.orbisgis.org/repository/orbisgis/')
@Grab(group='org.orbisgis.orbisprocess', module='geoclimate', version='1.0.0-RC1')

import org.orbisgis.orbisprocess.geoclimate.Geoclimate
Geoclimate.logger = logger

def process = Geoclimate.BDTOPO_V2.workflow()
process.execute(configurationFile:'/home/geoclimate/bdtopo/config_all_zones.json')
```

and same for OSM

Before

```groovy
...
def process = Geoclimate.Workflow.OSM()
...
```
After

```groovy
...
def process = Geoclimate.OSM.workflow()
...
```

Right ?

@j3r3m1

ebocher:
Yes but without parenthesis 
```
def process = Geoclimate.OSM.workflow
```
so you can do

```
Geoclimate.OSM.workflow.execute(configurationFile:'/home/geoclimate/bdtopo/config_all_zones.json')
```

What about if we proposed this signature for the workflow ?
Even simpler for users, isn't it?

```
Geoclimate.OSM.workflow.execute('/home/geoclimate/bdtopo/config_all_zones.json')
```

By the way test the syntax in groovy or dbeaver console. It works for me but in case of ;-)

gpetit:
Nice. Ok I will test and adapt the doc in order to have the simplest signature for users

gpetit:
I'm on it

gpetit:
The PR #335 partially answer this issue since only the new addresses of the configuration file have been updated.
[...]

PR #335, opened 2020-04-20
Update doc to manage new workflow oragnization

gpetit:
As asked in #332 , update the doc to take into account the new code organization, especially for the workflow and the configuration files

Correction of some typos


gpetit:
I updated the PR in order to remove the modifications about the workflow call.

So for the moment, we keep `Geoclimate.Workflow.BDTOPO_V2()` and `Geoclimate.Workflow.OSM()` that are still available for the RC1.

Once the RC2 will be released, I will update the doc in order to have instead `Geoclimate.BDTOPO_V2.workflow` and `Geoclimate.OSM.workflow`

PR #346, opened 2020-05-19
Remove the @Disabled in bdtopo_v2 tests

j3r3m1:
Remove the @Disabled in bdtopo_v2 tests (since we have a validated BDTopo dataset)


SPalominos:
The test fails because of the file path to BDTopo data. 
@j3r3m1 I've made you a fix through a PR

PR #354, opened 2020-05-29
Rework.

SPalominos:
This PR is a sample for the future reworking of Geoclimate coding syntax. The work is only done for the `geoindicators` module.

### Old behaviour
Currently the project is organized this way : 
- `Geoclimate` : root class of the project which contains the process categories (like `Geoindicators`, `Workflow`, `TypologyClassification` ...)
- `Geoindicators`, `ProcessingChain`, `BDTopo` : the process categories containing the process script (like `BlockIndicators`, `DataUtils` ...)
- `BlockIndicators`, `DataUtils` ... : the process script containing different processes (like `holeAreaDensity`, `closingness` ...)

So, to call a process you can do category.script.process (like `BuildGeoIndicators.Geoindicators.holeAreaDensity()`)

### New behaviour
In orbisdata/process-manager there is three concepts : 
- `Process` : a process with inputs and outputs
- `ProcessFactory` : a factory which simplify the process creation and contains `Process`
- `ProcessManager` : which simplify the `ProcessFactory` creation and contains `ProcessFactory`

So, to call a `Process` : `ProcessManager.ProcessFactory.Process`

The idea in this PR, is to avoid class declaraction like for `Geoindicators` and to use only Groovy script using`@BaseScript` annotation.

##### Process script
 So for the process script (`holeAreaDensity`, `closingness` ...), they should start with 
``` groovy
@BaseScript GroovyProcessFactory pf
```
With this annotation the script will be considered by Groovy as a `ProcessFactory`.
Then, to add the processes, you have to call the `create(Closure)` methods for each process : 
``` groovy
create {
    title "title of the process"
    id "identifier of the process"
    inputs in1: String, in2: ""
    outputs out1: String
    run { in1, in2 -> ... [out1: "toto"] }
}
```
/!\ Very important difference with the old behaviour, you HAVE to add the `id ...` line in your process creation and replace the `...` with the process id (like `createRSU`, `effectiveTerrainRoughnessLength`, `holeAreaDensity`, `closingness` ...). It's good to have one word id.

This way, when the `ProcessFactory` (process script) is loaded, the groovy script is executed once and all the processes are loaded and available (before the processes were created every time they were requested.) 

Using the annotation `@BaseScript ProcessFactory pf` also give access to several utility methods like :
- Logger methdos :
   - `info(Object)` and `info(Object, Exception)`
   - `warn(Object)` and `warn(Object, Exception)`
   - `error(Object)` and `error(Object, Exception)`
- pre/postfix methods : 
   - `prefix(String)` : add an UUID as prefix 
   - `prefix(String, String)` : use the first string as prefix of the second
   - `postfix(String)` : add an UUID as postfix 
   - `postfix(String, String)` : use the second string as postfix of the first

##### Process category
The process categories (`Geoindicators`, `ProcessingChain`, `BDTopo`) are much more simple.

They should starts with :
``` groovy
@BaseScript GroovyProcessManager pm
```
Then the `ProcessFactory` (process script)  should be register this way : 
``` groovy
register([ProcessFactoryClass1, ProcessFactoryClass2, ProcessFactoryClass3, ProcessFactoryClass4])
```

So when the `ProcessManager` (process category) is loaded, all the `ProcessFactory` with their `Process` are loaded.

To get a process you just have to first load the `ProcessManager` :
``` groovy
def PM = GroovyProcessManager.load MyProcessManager
PM.Factory.Process.execute([...])
```

@ebocher, @gpetit, @ELSW56, @j3r3m1 I hope I'm clear, if you have a question,just ask it on this PR. I'm not expecting you to make the full convertion of Geoclime, I'm on it. It's to inform you about the changes.


j3r3m1:
Not sure I really got the real need for this reworking. Is it to access more easily logger into groovy, to simplify the use of the processes inside the geoclimate code (for 'developers'), to simplify the use of the processes inside the groovy scripts (for 'users') ?
Could you give a 'real' (with an existing process) example of how:
- to define a new process in IntelliJ
- to use this process in a groovy script

Thx !

SPalominos:
[Here](https://github.com/orbisgis/geoclimate/blob/6e64627e5477ca93ed8499826cb6ebcf065634c9/geoindicators/src/main/groovy/org/orbisgis/orbisprocess/geoclimate/geoindicators/BlockIndicators.groovy)  and [here](https://github.com/orbisgis/geoclimate/blob/6e64627e5477ca93ed8499826cb6ebcf065634c9/geoindicators/src/main/groovy/org/orbisgis/orbisprocess/geoclimate/geoindicators/Geoindicators.groovy) you have a real example about how to define a new process.

The idea of the modification is to simplify the process declaration (for developpers).
But for final users, it won't change. Here an example of simple groovy script : 
``` groovy
//Grab the geoclimate dependency
@GrabResolver(name='orbisgis', root='https://nexus.orbisgis.org/repository/orbisgis/')
@Grab(group='org.orbisgis.orbisprocess', module='geoclimate', version='1.0.0-RC1')

//Import the Geoclimate root object
import org.orbisgis.orbisprocess.geoclimate.Geoclimate

//Get and run a process
def process = Geoclimate.Workflow.OSM
process(configurationFile:'/home/geoclimate/osm/osm_workflow_placename_folderoutput.json')
```

j3r3m1:
OK definitely much clearer than the old syntax. Well done

ebocher:
Doesn't compile.

SPalominos:
Normal, it needs the update of ObisData and OrbisAnalysis which are coming

PR #366, opened 2020-06-12
Add shortcut for workflow processes.

SPalominos:
Add shortcut for workflow processes.
Workflow are called this way : 
``` groovy
Geoclimate.OSM.Workflow
Geoclimate.BDTopo_V2.Workflow
```


ebocher:
Doesn't not compile

osm/src/main/groovy/org/orbisgis/orbisprocess/geoclimate/osm/WorkflowOSM.groovy
```diff
@@ -115,7 +115,7 @@ def ProcessingChain = GroovyProcessManager.load(PC)
  */
 create {
     title "Create all Geoindicators from OSM data"
-    id "workflow"
+    id "Workflow"
```
ebocher, line 118: I think it will be better to set the the first letter in lower for processing name

PR #367, opened 2020-06-12
Refactoring and fix workflows

ebocher:
This PR 

- merges all geoindicators chain in one process chain called GeoIndicatorsChain (see ProcessingChain module) 
To compute all indicators uses the syntax

```java
ProcessingChain.GeoIndicatorsChain.computeAllGeoIndicators
```

- BTopo workflow uses now the same method as OSM workflow to compute the geoindicators

- Impervious surface is now also used in all workflow

-  adds a surface criterion for filtering RSU too small. default is 0.1

There are still a number of mysteries with the new factory system.

Sometimes the processes are found, sometimes they're empty.

It's a little complicated to understand. @SPalominos  will have a better point of view.


osm/src/main/groovy/org/orbisgis/orbisprocess/geoclimate/osm/WorkflowOSM.groovy
```diff
@@ -419,11 +419,12 @@ create {
                             info "OSM GIS layers formated"
 
                             //Build the indicators
-                            IProcess geoIndicators = ProcessingChain.Workflow.GeoIndicators
+                            IProcess geoIndicators = ProcessingChain.GeoIndicatorsChain.computeAllGeoIndicators
```
j3r3m1, line 422: well done

SPalominos:
@ebocher I did to your geoclimate-1 repo a PR fixing the merge and fixing some tests.

ebocher:
Thanks

PR #380, opened 2020-06-22
Syntax update

SPalominos:
An update for the geoclimate project syntax in order to make IDE autocompletion and navigation available again. More commit are comming, but first @ebocher and @j3r3m1 tell me what you think about the syntax. Before merge, a PR on OrbisData is comming.

The main change is the usage of the `Field` annotation which allow to use only script and no classes.

In the `ProcessManager` (here in the file `BDTopo_V2.groovy`), the declaration of a factory (like `AbstractTablesInitialization.groovy `) is done with two lines :
``` groovy
//Make the factory visible by the IDE
@Field static AbstractTablesInitialization abstractTablesInitialization
//Register the factory
abstractTablesInitialization = register(AbstractTablesInitialization)
```

In the `ProcessFactory`, a field with the process id as name is declared :
``` groovy
@Field IProcess prepareData = create { ... }
```

Now to call a process you can write :
``` groovy
//If only importing Geoclimate
def p = Geoclimate.BDTopo_V2.AbstractTablesInitialization.initParametersAbstract
//Or
def p = BDTopo.PrepareBDTopo.prepareData
```

### DIFF
--- a/bdtopo_v2/src/main/groovy/org/orbisgis/orbisprocess/geoclimate/bdtopo_v2/AbstractTablesInitialization.groovy
+++ b/bdtopo_v2/src/main/groovy/org/orbisgis/orbisprocess/geoclimate/bdtopo_v2/AbstractTablesInitialization.groovy
@@ -1,7 +1,9 @@
 package org.orbisgis.orbisprocess.geoclimate.bdtopo_v2
 
 import groovy.transform.BaseScript
+import groovy.transform.Field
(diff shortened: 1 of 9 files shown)


j3r3m1:
> ```groovy
> @Field IProcess prepareData = create { ... }
> ```
I am not sure I got this one.

SPalominos:
Ok, reverting to the old syntax

PR #387, opened 2020-06-25
New lcz test

j3r3m1:
Fix some confusions observed between compact mid-rise (5) and compact low-rise (6).
The problem was due to a problem of LCZ limitations defined in the code.


j3r3m1:
I can not reproduce the error on my machine, I have tried restarted the test. @SPalominos @ebocher, an idea where the problem could come from ?

ebocher:
I'm on it

ebocher:
Yessssssssssssssss

Issue #402, opened 2020-08-05
Error message wrong when configFile does not exist


j3r3m1:
When the workflow does not find the configuration file, the following error message is displayed:
> [ERROR] org.orbisgis.orbisdata.processmanager.process.GroovyProcessFactory - Invalid file parameters

All my parameters were valid but the path of my configuration file was wrong. Then it could be much more consistent to display an error message such as:
"Configuration file not found"

ELSW56:
See PR #403

PR #403, opened 2020-08-24
Modifications to take issue #402 into account

ELSW56:
Error message modified if parameters file is not found


ebocher:
wunderbar !

PR #405, opened 2020-08-24
Add new function to gather all scales in one table in order to use ra…

j3r3m1:
…ndom forest algorithms (at RSU or building scale)

Add the RF algorithm in the geoindicator function for LCZ calculation


j3r3m1:
Do not merge, not terminated yet

PR #408, opened 2020-09-07
Addition of a parameter estimateHeight

ELSW56:
If true, an additional table, with the same name as the building table preceded by 'EST_', will be created together with the buildings layer, giving the status of its height in the column estimated.

If estimated = true, it means that the corresponding building height should be estimated thanks to the classification.


ebocher (approved):
Brilliant !

PR #416, opened 2020-09-28
Zonal area

ernlt:


ebocher:
Merged in #417 with some fixes

PR #424, opened 2020-10-09
Fix lcz equality val

j3r3m1:
There was an issue with the LCZ classification for some areas. The problem came from the fact that some buildings were considered as RSU, thus having no RSU aspect ratio (= null). The distrib characterization has been modified in order to set these RSU at "unknown". Moreover, for the same geometries, the SVF was set by default to 1 while it should be lower than 0.5. Thus we make it work for this specific areas


ebocher:
championnnnnnnnnnnnnnnnnn

PR #452, opened 2020-10-21
set lcz to 0 when no more than 50% of the ground is covered by pervio…

j3r3m1:
…us or impervious fraction

Several zones did not have any information about the land cover. Initially we consider that the largest rate is the one of the RSU even though the rate is 1% (except for the impervious which was 5%).
This solution was quite good in general. However, it could lead to very inappropriate LCZ in certain cases and these cases are not easy to identify. Thus we have decide to set an unknown category where less than 50% of the RSU is not covered by a land cover (different than building).


j3r3m1:
Would need comparison with the current version before to merge (especially for the BDTopo case). If there is too much difference we may set the unknow type in the second LCZ. This would lead to nice LCZ map and it would remain possible to identify rsu that are potentially erroneous

Issue #461, opened 2020-11-02
OSM exclude highway=service to compute RSU


ebocher:
It'd be nice to exclude the highway=service tag to the RSU input data to avoid small areas.

ebocher:
Fixed

Issue #414, opened 2020-09-18
BD Topo failed for insee code 31160 (Daux)


j3r3m1:
This commune fails when executing the identification of the minimum geometry for the surface fraction calculation. Here is the error message:
> [INFO] org.orbisgis.orbisdata.processmanager.process.GroovyProcessFactory - Generating minimum polygon areas
[ERROR] org.orbisgis.orbisdata.processmanager.process.Process - Error while executing the process.
org.h2.jdbc.JdbcSQLNonTransientException: Exception lors de l'appel de la fonction définie par l'utilisateur: "union(MULTILINESTRING ((559472.194 6291021.554, 559472.194 6291021.555), (559472.194 6291021.555, 559469.1 6291031.1), (559469.1 6291031.1, 559466.631 6291037.739), (559466.631 6291037.739, 559469.5 6291038.6), (559469.5 6291038.6, 559470.5 6291039.6), (559470.5 6291039.6, 559471.6 6291046.6), (559471.6 6291046.6, 559.....................
 558873.936 6292657.327, 558867.537 6292593.066), (558867.537 6292593.066, 558864.775 6292593.087), (558864.775 6292593.087, 558871.2 6292657.6), (558871.2 6292657.6, 558871.525 6292662.192), (558871.525 6292662.192, 558874.301 6292662.47))): found non-noded intersection between LINESTRING ( 558852.2752000159 6292093.324799984, 558851.9990000001 6292093.601 ) and LINESTRING ( 558852.369 6292093.231, 558851.999 6292093.601 ) [ (558852.219340668, 6292093.380659332, NaN) ]"; SQL statement:
(select st_polygonize(st_union(
                st_precisionreducer(st_node(st_accum(st_force2d(a.the_geom))), 3))) as the_geom, id_rsu from tmp_tables_zindex0_593e5db7_bdf5_4352_b2e9_63ac3a51be78 as a group by id_rsu) [90105-200]

ebocher:
Fixed. See the beautiful LCZ map
![daux_lcz](https://user-images.githubusercontent.com/1820572/98704916-7a9e2200-237d-11eb-83e9-61508b6306c6.png)

ebocher:
Fixed

PR #464, opened 2020-11-10
Last but not least fixes!

ebocher:
About
#414 , #439 , #461 
Plus a fixe on rural LCZ type


bdtopo_v2/src/main/groovy/org/orbisgis/orbisprocess/geoclimate/bdtopo_v2/WorkflowBDTopo_V2.groovy
```diff
@@ -933,7 +933,7 @@ def extractProcessingParameters(def processing_parameters){
     def defaultParameters = [distance: 1000,indicatorUse: ["LCZ", "URBAN_TYPOLOGY", "TEB"],
                              svfSimplified:false, prefixName: "",
                              mapOfWeights : ["sky_view_factor"                : 4,
```
j3r3m1, line 935: I wonder whether in the OSM workflow (not here) the weights for "height of roughness" and "SVF" should be inverted (the first have 4 and the second have 6) because the SVF rely on building height but also building distribution while the height of roughness only rely on bulding height (which is roughly estimated...)

osm/src/main/groovy/org/orbisgis/orbisprocess/geoclimate/osm/WorkflowOSM.groovy
```diff
@@ -953,7 +953,7 @@ def extractProcessingParameters(def processing_parameters){
     def defaultParameters = [distance: 0,indicatorUse: ["LCZ", "URBAN_TYPOLOGY", "TEB"],
                              svfSimplified:false, prefixName: "",
                              mapOfWeights :  ["sky_view_factor"                : 4,
```
j3r3m1, line 955: previous comment apply here

j3r3m1 (approved):

geoindicators/src/main/groovy/org/orbisgis/orbisprocess/geoclimate/geoindicators/TypologyClassification.groovy
```diff
@@ -189,7 +189,7 @@ IProcess identifyLczType() {
                                         AS SELECT   $ID_FIELD_RSU,
                                                 CASE WHEN IMPERVIOUS_FRACTION_LCZ>PERVIOUS_FRACTION_LCZ AND IMPERVIOUS_FRACTION_LCZ>0.05
                                                         THEN 105
-                                                        ELSE CASE WHEN ALL_VEGETATION<WATER_FRACTION_LCZ OR WATER_FRACTION_LCZ > 0.5
+                                                        ELSE CASE WHEN (ALL_VEGETATION<WATER_FRACTION_LCZ AND WATER_FRACTION_LCZ> 0.5)  OR WATER_FRACTION_LCZ > 0.5
```
j3r3m1, line 192: I think that "OR WATER_FRACTION_LCZ > 0.5" is not useful because redundant and remove the condition "ALL_VEGETATION<WATER_FRACTION_LCZ".
Moreover I am actually not sure about the adding of WATER_FRACTION_LCZ>0.5 because you may have:
building_fraction=0.09
Impervious=0.3
all_vegetation=0.3
water_fraction_lcz=0.31
The sum of all fractions is still 1 but the major part is water. So maybe >0.31 would make more sense ?

geoindicators/src/main/groovy/org/orbisgis/orbisprocess/geoclimate/geoindicators/TypologyClassification.groovy
```diff
@@ -189,7 +189,7 @@ IProcess identifyLczType() {
                                         AS SELECT   $ID_FIELD_RSU,
                                                 CASE WHEN IMPERVIOUS_FRACTION_LCZ>PERVIOUS_FRACTION_LCZ AND IMPERVIOUS_FRACTION_LCZ>0.05
```
j3r3m1, line 190: it is necessary to add some conditions here, other wise the impervious_fraction may be larger than water and than vegetation but it will not be affected with impervious. Here is the replacement piece of query:
CASE WHEN IMPERVIOUS_FRACTION_LCZ>ALL_VEGETATION AND IMPERVIOUS_FRACTION_LCZ>WATER_FRACTION_LCZ AND IMPERVIOUS_FRACTION_LCZ>0.05

j3r3m1 (approved):

osm/src/main/groovy/org/orbisgis/orbisprocess/geoclimate/osm/WorkflowOSM.groovy
ebocher: it makes sense

geoindicators/src/main/groovy/org/orbisgis/orbisprocess/geoclimate/geoindicators/TypologyClassification.groovy
ebocher: It's not redundant because you parenthesis precedence.

 The sum of all fractions is still 1 but the major part is water. So maybe >0.31 would make more sense ?
Yes

PR #481, opened 2020-11-23
Lcz improvement

j3r3m1:
The objective of this PR is:
- to deal with LCZ1 issue: in a lot of cases we have LCZ type 1 having less than 10 stories. This is solved by this issue
- to deal with the fact that LCZ8 are under-represented. We set a specific rule to have LCZ8 when the commercial fraction > residential fraction and > industrial fraction and >0.33 and nb_lev < 3


geoindicators/src/main/groovy/org/orbisgis/orbisprocess/geoclimate/geoindicators/GenericIndicators.groovy
```diff
@@ -561,6 +564,9 @@ IProcess distributionCharacterization() {
                             def id_rsu = rowMap."$inputId"
                             rowMap.remove(inputId.toUpperCase())
                             def sortedMap = rowMap.sort { it.value }
+                            // We want to get rid of some of the values identified as -9999.99
+                            while (sortedMap.values().remove(-9999.99 as double));
+                            println sortedMap.keySet()[idxExtrem]
```
ebocher, line 569: remove println

geoindicators/src/main/groovy/org/orbisgis/orbisprocess/geoclimate/geoindicators/GenericIndicators.groovy
```diff
@@ -523,6 +523,9 @@ IProcess distributionCharacterization() {
...
```
ebocher, line 528: remove println

Issue #480, opened 2020-11-22
New building Type for LCZ need


j3r3m1:
A new way to get LCZ8 (large low-rise) is to use (between others) the fraction_commercial>0.5. However, LCZ8 may be commercial or light industrial (such as industrial warehousing). In order to differentiate LCZ8 to LCZ10 (heavy industry), it may be interesting to create a type for warehouse (if we do have any in OSM or BDTopo).

ebocher:
There is a tag in OSM https://wiki.openstreetmap.org/wiki/Tag:building%3Dwarehouse
but the tag is often not filled. 
What about setting the type warehouse when the building is in a landuse=industrial. It seems the good practice for OSM https://wiki.openstreetmap.org/wiki/Tag:industrial%3Dwarehouse

By the way even if we set the type warehouse. I don't think it will be enough to separate  light industrial from heavy industry

j3r3m1:
I would say that most warehouse are light industrial, it is more storage than hard process consuming tons of fuel.

j3r3m1:
OK it seems necessary to distinguish light industry and heavy industry.

- By default if there is no information more than industrial, let's consider it is light industry.
- Then it is necessary to tag as heavy industry buildings or zone such as defined in Stewart (2011):
```LCZ 10 is the site of major industrial facilities such as factories, refineries, mills,
and processing plants. Its buildings release large quantities of atmospheric pollutants and waste heat from mechanical and chemical processing, which drives QF above 300 W m-2. Buildings and other structures (tanks, towers, stacks) are non-uniform in size, height, and distribution

@ebocher can do that in OSM
@gpetit any chance you would have time to do it for BDTopo or it is better we do it ?

ebocher:
I'm on it for OSM part.

ebocher:
Ok done for BDTopo using the RESERVOIR table. Merged with building see https://github.com/orbisgis/geoclimate/pull/497
Need some update with the LCZ classification due to the new type heavy_industry and light_industry

ebocher:
Done

PR #497, opened 2020-12-14
Extract heavy and light industry from OSM and BDTopo

ebocher:
Add new building types : heavy_industry and light_industry have been added from BDTopo and OSM datasources to qualify the building.
Note : light_industry is now the default type instead of industry

About #480


ebocher:
LCZ classification must be updated regarding the two new industry types
This a job for a champion !
@j3r3m1

Issue #509, opened 2021-01-13
Only building height estimation


j3r3m1:
In the new version of the configuration file, we need to set at least one "indicatorUse" type (LCZ, TEB, URBAN_TYPO). However, for OSM, a user could only want to get OSM data with estimated building height.
Thus it would be nice if the workflow could work with "indicatorUse" = [], ie calculate indicators for building height estimation and then nothing else.

ebocher:
Good idea

ebocher:
Done

PR #511, opened 2021-01-14
Add a way to estimated the building height without re-computing indicators

ebocher:
About https://github.com/orbisgis/geoclimate/issues/509


j3r3m1 (approved):
I did not check but did you update the input building table with building height in previous PR ?

PR #516, opened 2021-01-22
Ascii grid output and temporary tables deletion

ELSW56:
WorkflowOSM.groovy : Adding of a function to save the results as ascii grid files.

DataUtils.groovy : Adding of a function to drop the intermediate tables (applied only to grid_indicators in GeoIndicatorsChain.groovy up to now)

The changes in the ProcessingChainOSMTests.groovy file are just due to indentation differences ...


ebocher:
Thanks I'm on it

PR #517, opened 2021-01-25
Add a method to store the grid indicators in esri ascii grid format

ebocher:
This PR is based on https://github.com/orbisgis/geoclimate/pull/516
It introduces a method to save the grid indicators in ascii grid.
The srid is detected to reproject the grid if necessary
A prj file is written to save the prj informations
A unit test is added to check if the grid is created

Thanks to @ELSW56

### DIFF
--- a/osm/src/main/groovy/org/orbisgis/orbisprocess/geoclimate/osm/WorkflowOSM.groovy
+++ b/osm/src/main/groovy/org/orbisgis/orbisprocess/geoclimate/osm/WorkflowOSM.groovy
@@ -18,6 +18,7 @@ import org.orbisgis.orbisdata.datamanager.jdbc.postgis.POSTGIS
 import org.orbisgis.orbisdata.processmanager.api.IProcess
 import org.orbisgis.orbisprocess.geoclimate.processingchain.ProcessingChain
 import org.orbisgis.orbisanalysis.osm.OSMTools
+import org.h2gis.functions.io.utility.PRJUtil;
 
 import java.sql.Connection
 import java.sql.SQLException
@@ -554,6 +555,7 @@ IProcess osm_processing() {
 
                             //Add the GIS layers to the list of results
                             def results = [:]
+                            results.put("outputTableZone", zoneTableName)
                             results.put("roadTableName", roadTableName)
                             results.put("railTableName", railTableName)
                             results.put("hydrographicTableName", hydrographicTableName)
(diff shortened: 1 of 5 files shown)


osm/src/main/groovy/org/orbisgis/orbisprocess/geoclimate/osm/WorkflowOSM.groovy
```diff
@@ -588,11 +590,10 @@ IProcess osm_processing() {
                                      results.putAll(geoIndicators.getResults())
                                 }
                             }
-
+                            def x_size = grid_indicators_params.x_size
+                            def y_size = grid_indicators_params.y_size
```
ELSW56, line 594: Why do you define x_size and y_size before verifying that grid_indicators_params is set ?

osm/src/main/groovy/org/orbisgis/orbisprocess/geoclimate/osm/WorkflowOSM.groovy
```diff
@@ -1091,14 +1095,86 @@ def saveOutputFiles(def h2gis_datasource, def id_zone, def results, def outputFi
             saveTableAsGeojson(results.outputTableBuildingUrbanTypo, "${subFolder.getAbsolutePath()+File.separator+"building_urban_typo"}.geojson", h2gis_datasource,outputSRID,reproject,deleteOutputData)
         }
         else if(it.equals("grid_indicators")){
-            saveTableAsGeojson(results.grid_indicators, "${subFolder.getAbsolutePath()+File.separator+"grid_indicators"}.geojson", h2gis_datasource,outputSRID,reproject,deleteOutputData)
+            if(x_size!=y_size){
```
ELSW56, line 1098: Shouldn't the user be involved in choosing the output format he would like when the choice exists ?

osm/src/main/groovy/org/orbisgis/orbisprocess/geoclimate/osm/WorkflowOSM.groovy
ebocher: to avoid null variable in the save method

osm/src/main/groovy/org/orbisgis/orbisprocess/geoclimate/osm/WorkflowOSM.groovy
ebocher: This is a good question and not easy to solve because there is have no parameter in the config file to define the output file format.

osm/src/main/groovy/org/orbisgis/orbisprocess/geoclimate/osm/WorkflowOSM.groovy
ELSW56: OK, but what happens if grid_indicators_params is not set ?

ebocher: oupps. Good catch

ELSW56: I suggest this instead of lines 593 to 596 by
```
                            def x_size
                            def y_size
                            //Compute the grid indicators
                            if(grid_indicators_params){
                                 x_size = grid_indicators_params.x_size
                                 y_size = grid_indicators_params.y_size
```

Then, in saveOutputFiles at line 1098, the condition should be modified as :
```
            if(!xsize ||!ysize || x_size!=y_size){
```
instead of 
```
            if(x_size!=y_size){
```
What do you think of this, @ebocher ?

PR #521, opened 2021-01-26
Fix SRID type

ELSW56:
Fix SRID type in WorkFlowOSM


ebocher:
Merged with some changes on BDTOPO thanks

Issue #550, opened 2021-03-24
Chain crashing for some areas during application of the building_height_osm_rf_2_0.model


ELSW56:
It happens several time: I'll give the other bboxes concerned in further comments.
Here's the errors stack trace associated for the zone [53.41396, 5.33682, 53.49729, 5.42015]:
```
All geoindicators have been computed
Extracting the building having no height information and estimate it
Collect building indicators to estimate the height
Start estimating the building height
Security framework of XStream not explicitly initialized, using predefined black list on your own risk.
Error while executing the process.
java.lang.IndexOutOfBoundsException: Index 0 out of bounds for length 0
	at jdk.internal.util.Preconditions.outOfBounds(Preconditions.java:64) ~[?:?]
	at jdk.internal.util.Preconditions.outOfBoundsCheckIndex(Preconditions.java:70) ~[?:?]
	at jdk.internal.util.Preconditions.checkIndex(Preconditions.java:248) ~[?:?]
	at java.util.Objects.checkIndex(Objects.java:372) ~[?:?]
	at java.util.ArrayList.get(ArrayList.java:458) ~[?:?]
	at smile.data.DataFrameImpl.<init>(DataFrameImpl.java:226) ~[geoclimate-1.0.0-SNAPSHOT-jar-with-dependencies.jar:?]
	at smile.data.DataFrame.of(DataFrame.java:1230) ~[geoclimate-1.0.0-SNAPSHOT-jar-with-dependencies.jar:?]
	at org.orbisgis.orbisdata.datamanager.dataframe.DataFrame.of(DataFrame.java:1019) ~[geoclimate-1.0.0-SNAPSHOT-jar-with-dependencies.jar:?]
	at org.orbisgis.orbisdata.datamanager.dataframe.DataFrame$of.call(Unknown Source) ~[?:?]
	at org.orbisgis.orbisprocess.geoclimate.geoindicators.TypologyClassification$_applyRandomForestModel_closure3$_closure11.doCall(TypologyClassification.groovy:664) ~[geoclimate-1.0.0-SNAPSHOT-jar-with-dependencies.jar:?]
	at jdk.internal.reflect.GeneratedMethodAccessor553.invoke(Unknown Source) ~[?:?]
	at jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43) ~[?:?]
	at java.lang.reflect.Method.invoke(Method.java:566) ~[?:?]
[...]

PR #553, opened 2021-03-24
Solves building height RF error when empty input

j3r3m1:
Avoid the RF for building height estimation when there is no building height to estimate.
Solves #550


ebocher:
Brilliant !

PR #584, opened 2021-06-07
Change module architecture

ebocher:
This PR  concerns a major redesign of the GeoClimate modules to be inline with this description https://github.com/orbisgis/geoclimate/wiki/Coding-implementation


ebocher:
more investigations are needed to support Map with named parameters

j3r3m1 (approved):
Approve

Issue #557, opened 2021-04-15
Change Urban Typology name in the code


j3r3m1:
We renamed Urban typology by UTRF (Urban Typology by Random Forest) in the documentation but it is now not consistent anymore with the code (for example in the config file indicatorUse is URBAN_TYPOLOGY. We should update this term.

Note that it will also impact the indicator names (those ending by URB --> UTRF). For now they are still URB in the wiki, need to be updated when done in the code

ebocher:
So you want to rename the indicatorUse variable to  UTRF or the full name Urban_Typology_by_Random_Forest
Note that the output tables must renamed too

j3r3m1:
Yes. I can do that if you want

ebocher:
Do it ;-)
Note you must change it also in the BDTopo worflow

j3r3m1:
When #590 merged, remains the doc to update

ebocher:
Done

PR #590, opened 2021-06-15
Update utrf name

j3r3m1:
Modifications according to #557 
Remains doc to update


ebocher:
In BDTopo and OSM workflows the saved tables must be updated also https://github.com/orbisgis/geoclimate/blob/6a00173c420f06157df4622041a9f504e214859d/osm/src/main/groovy/org/orbisgis/geoclimate/osm/WorkflowOSM.groovy#L824
What about rsu_urtf_area ?

j3r3m1:
> In BDTopo and OSM workflows the saved tables must be updated also
> 
> https://github.com/orbisgis/geoclimate/blob/6a00173c420f06157df4622041a9f504e214859d/osm/src/main/groovy/org/orbisgis/geoclimate/osm/WorkflowOSM.groovy#L824
> 
> What about rsu_urtf_area ?

There are a lot of variables or names using "urban_typo" and I though only solving the ones at "top" scale would be sufficient

ebocher:
ok

ebocher:
Fixed

PR #593, opened 2021-06-17
Clean main geoclimate module + remove old doc

ebocher:


geoclimate/src/main/groovy/org/orbisgis/geoclimate/Geoclimate.groovy
```diff
@@ -1,7 +1,5 @@
 package org.orbisgis.geoclimate
 
-import org.orbisgis.geoclimate.bdtopo_v2.BDTopo_V2
-import org.orbisgis.geoclimate.osm.OSM
```
j3r3m1, line 4: why do you remove this line ? OSM is not declared anymore then right ?

j3r3m1 (changes requested):

j3r3m1 (approved):

PR #610, opened 2021-07-15
Update according to openjournals/joss-reviews#3442

j3r3m1:
Remove all text from the reference section
Put all footnotes which were in the reference section within the article body


arfon:
@j3r3m1 - references need to be included without the back-ticks. i.e. like this [@masson2000] not `[@masson2000]`

PR #618, opened 2021-08-27
Update condition on output SRID

j3r3m1:
Small fix to make work the output srid key word...


ebocher:
Good catch
outputsrid can be null and greater to zero

ebocher:
Added in https://github.com/orbisgis/geoclimate/pull/620

PR #632, opened 2021-09-23
add screenshot 2

franetibe:
New screenshots for the wiki (windows part).


j3r3m1 (commented):
Good job !
Are all files used in the wiki (you added 8 figures) or some of them should be removed ?

ebocher:
Thanks a lot!
Some files seem redundant.  For example cli and groovyconsole are the same for Windows and Unix.

franetibe:
I just uploaded 8 figures, any removal needed.

Four pictures can be considered as redundant, however it is not strictly identical between linux and windows : 
- the CLI is purple with linux and black with windows
- the color window of the groovy console change between linux and windows
- the window is different between linux and windows for results_osm
- the window is different between linux and windows for results_bdtopov2

j3r3m1 (approved):

j3r3m1:
There is an error at the compilation with the OSM data.

`java.io.FileNotFoundException: /home/runner/work/geoclimate/geoclimate/osm/./target/geoclimate_chain_srid/osm_Pont-de-Veyle`

 @ebocher is it related to the fact that the target is empty and that the osm data need to be redownloaded on the server ?

PR #690, opened 2022-03-03
Add density and sum of population by RSU

ebocher:
Compute the density and the sum of population by RSU from the pop indicator available at building scale


j3r3m1 (approved):
what a change !

PR #700, opened 2022-03-11
Fix UTRF name to be aligned with the documentation

ebocher:
This PR change all urban typology references to the new term UTRF.
These changes required to update all training data and the RF models : URTF and OSM Building

about #590


ebocher:
This PR cannot compile with github action because it uses a model not yet deployed.
So I merge it

Issue #704, opened 2022-03-18
WRF indicators


j3r3m1:
To run WRF, the following urban characteristics should be calculated (according to [Chen et al. 2022](https://www.frontiersin.org/articles/10.3389/fclim.2021.771441/full), Table 1):

- [x] Mean building height: already calculated and rasterized ("BUILDING_HEIGHT")
- [x] Distribution of building heights: has to be done since slightly different from the "roofAreaDistribution()" Process
- [x] Area weighted mean building height: calculated (weightedAggregatedStatistics() Process) but not rasterized, should be rasterized using the same method as "BUILDING_HEIGHT")
- [x] Standard deviation of building height: already calculated and rasterized ("BUILDING_HEIGHT")
- [x] Plan area fraction: already calculated and rasterized ("BUILDING_FRACTION")
- [x] Building surface to plan area ratio: simple sum of two already calculated and rasterized indicators ("BUILDING_FRACTION" and "FREE_EXTERNAL_FACADE_DENSITY") -> a new indicator (and IProcess) should be created "BUILDING_SURFACE_FRACTION"
- [x] Frontal area index: the same as the "projectedFacadeAreaDistribution()" Process but when buildings are cutted by the cells... It seems this indicator is a fraction of the plan area and that it should also be divided by the height of the bin.

Additionnally, each pixel should be classified into one of the following three urban land use types:
- [x] Low-density residential (urban area with pop density < 10 people/ha), high-density residential (urban area with pop density included between 10 and 100 people/ha), commercial (urban area with pop density > 100 people/ha)

Each pixel should also have the fraction of land and sea:
- [x] #805

j3r3m1:
For persons willing to use GeoClimate to produce the WRF spatial inputs, they should add the following indicators to the list of indicators to calculate at grid scale:

> * Mean building height -> "BUILDING_HEIGHT"
> * Distribution of building heights -> "BUILDING_HEIGHT_DIST"
> * Area weighted mean building height -> "BUILDING_HEIGHT_WEIGHTED"
> * Standard deviation of building height ->  std is automatically calculated along with mean if you add "BUILDING_HEIGHT"
> * Plan area fraction -> "BUILDING_FRACTION"
> * Building surface to plan area ratio -> "BUILDING_SURFACE_DENSITY"
> * Frontal area index -> "FRONTAL_AREA_INDEX"
> * fraction of sea/land -> "SEA_LAND_FRACTION"

[...]

PR #705, opened 2022-03-18
Wrf indicators

j3r3m1:
According to issue #704


geoindicators/src/main/groovy/org/orbisgis/geoclimate/geoindicators/WorkflowGeoIndicators.groovy
```diff
@@ -2119,12 +2127,126 @@ IProcess rasterizeIndicators() {
...
+                    } else {
+                        info "Cannot compute the frontal area index."
+                    }
+                }
+
+                if(list_indicators*.toUpperCase().contains("SEA_LAND_FRACTION") && seaLandMaskTableName) {
+                    // If only one type of surface (land or sea) is in the zone, no need for computational fraction calculation
+                    def nbTypes = datasource.firstRow("""SELECT COUNT(DISTINCT($seaLandTypeField)) AS nb_types FROM $seaLandMaskTableName""").NB_TYPES
+                    if (nbTypes == 1) {
+                        def uniqueSurfaceType = datasource.firstRow("""SELECT DISTINCT($seaLandTypeField) AS unique_surface_type FROM $seaLandMaskTableName""").UNIQUE_SURFACE_TYPE
```
ebocher, line 2214: You can reduce to one query with

rows = datasource.rows(SELECT COUNT(*) AS nb_types FROM $seaLandMaskTableName group by ($seaLandTypeField)"""))

and then use the rows to check if there is one seaLandTypeField equals to sea

ebocher (commented):
So brilliant ;-)

PR #713, opened 2022-03-23
Updated 'lcz_primary' and 'lcz_secondary' style files

maxcollombin:
Correction a small inconsistency in the definition of the style to be applied for the primary and secondary lcz (see picture)

(It might also be interesting to reference these files in the wiki so that Geoclimate users can download them directly).

image


ebocher:
Thanks a lot @maxcollombin 
Good idea about a wiki page with a set of styles available

maxcollombin:
Yes, absolutely @j3r3m1.
The fact is that I first had to recreate a style file in qgis when you already had some available.

j3r3m1:
OK done for LCZ and UTRF, thanks for the contribution !

Issue #715, opened 2022-03-28
Inconsistencies between style file and wiki


maxcollombin:
Some small inconsistencies were found in the urban_typo style file on lines:

- https://github.com/orbisgis/geoclimate/blob/master/geoindicators/src/main/resources/styles/urban_typo.sld#L55-L58
- https://github.com/orbisgis/geoclimate/blob/master/geoindicators/src/main/resources/styles/urban_typo.sld#L78-L81
- https://github.com/orbisgis/geoclimate/blob/master/geoindicators/src/main/resources/styles/urban_typo.sld#L147-L150
- https://github.com/orbisgis/geoclimate/blob/master/geoindicators/src/main/resources/styles/urban_typo.sld#L170-L173

 and related information in the wiki:

https://github.com/orbisgis/geoclimate/wiki/UTRF-classification

It might also be interesting to keep some consistency between the names of the data files and those of the associated style files (ex urban_typo.sld, rsu_urban_typo_area.geojson, rsu_urban_typo_floor_area.geojson).

I can submit a PR if necessary...

ebocher:
Hey @maxcollombin 
You are welcome to send a PR.
Thanks a lot

PR #716, opened 2022-03-28
Small update on the style files

maxcollombin:
- Small update on the style files according to #715 
- Added the French version of the style files


ebocher:
Thanks a lot. Could you please change EN and FR to lowercase

PR #751, opened 2022-07-08
Modify workflow grid

j3r3m1:
There was two mistakes in the workflow:
- troubles to calculate sea/land fractions in the rasterize function
- values written in the ascii grid were null instead of -9999


geoindicators/src/main/groovy/org/orbisgis/geoclimate/geoindicators/WorkflowGeoIndicators.groovy
```diff
@@ -2216,26 +2216,30 @@ IProcess rasterizeIndicators() {
...
-                    datasource """ 
+                    datasource """
                             DROP TABLE IF EXISTS $tesselatedSeaLandTab;
                             CREATE TABLE $tesselatedSeaLandTab(id_tesselate serial, the_geom geometry, $seaLandTypeField VARCHAR)
-                                AS SELECT null, the_geom, $seaLandTypeField
-                                FROM ST_EXPLODE('select st_tesselate(the_geom) AS the_geom, $seaLandTypeField FROM $seaLandMaskTableName')"""
+                                AS SELECT explod_id, the_geom, $seaLandTypeField
+                                FROM ST_EXPLODE('(SELECT st_tessellate(the_geom) AS the_geom, $seaLandTypeField 
+                                                    FROM $seaLandMaskTableName
+                                                    WHERE ST_DIMENSION(the_geom) = 2 AND ST_ISEMPTY(the_geom) IS NOT TRUE
```
j3r3m1, line 2225: I suppose we can directly do that when the sea_land_mask is created ?

ebocher (approved):
Exceptionnel, brillant !

PR #762, opened 2022-09-09
Introduce several changes on geoclimate

ebocher:
- do not use highway= pedestrian and area=yes in the RSU process. This area must be considered in impervious layer.
- merge all tables one by one when the area is defined with a multipolygon, to obtain a continous domain
- use postfix instead of prefix to set a unique name for the output results. Indeed sometimes prefix is empty
- upgrade xstream lib

### DIFF
--- a/bdtopo_v2/src/main/groovy/org/orbisgis/geoclimate/bdtopo_v2/InputDataFormatting.groovy
+++ b/bdtopo_v2/src/main/groovy/org/orbisgis/geoclimate/bdtopo_v2/InputDataFormatting.groovy
@@ -41,7 +41,7 @@ IProcess formatBuildingLayer() {
                 hThresholdLev2 = 10
             }
 
-            def outputTableName = postfix "INPUT_BUILDING"
+            def outputTableName = postfix "BUILDING"
             debug 'Formating building layer'
 
             //Create the final building table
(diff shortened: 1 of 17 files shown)


j3r3m1:
>     * use postfix instead of prefix to set a unique name for the output results. Indeed sometimes prefix is empty
Is this really useful ? The prefix is used to distinguish two calculations. Using postfix makes it difficult to debug sometimes since it is difficult to know which is the last one. Moreover, it could lead to enormous databases if the database is not deleted regularly.

ebocher:
I' d like to said. Absolutely yes ,-)
Imagine you are running a workflow on two locations. Currently and for a few tables, the name of the ouput table is the same, this is the case for INPUT_ROAD. So you can't keep in the H2GIS database, the input_road table for both locations. Only the last one will be stored. This PR fixes this issue for all all output tables computed by the workflows.

ebocher:
Yes this will require some code and tests. 
but this does not solve the problem of tables with the same names. Or the user has to put a prefix by himself, which is very heavy.

ebocher:
Note that you can easly get the name of tables when a workflow is finished
e.g  process.getResults().output["Pont-de-Veyle"]["roadTrafficTableName"]

j3r3m1:
Bonjour,
    Je suis actuellement absent du bureau jusqu'au 15 octobre.
    Cordialement,
    Jérémy
    --
    Hi,
    I am actually out of office up to October the 15th.
    Best regards,
    Jérémy

PR #769, opened 2022-09-27
OSM workflow restructuring

ebocher:
This PR introduces improvements on OSM module :
- better code organization
- better OSM data building
- better management for table names


geoindicators/src/main/groovy/org/orbisgis/geoclimate/geoindicators/BuildingIndicators.groovy
```diff
@@ -162,8 +162,8 @@ IProcess neighborsProperties() {
             }
             query += """$ID_FIELD FROM (
                 SELECT 
-                    ST_INTERSECTION(a.$GEOMETRIC_FIELD,
-                    b.$GEOMETRIC_FIELD) AS the_geom,
+                    ST_INTERSECTION(st_makevalid(a.$GEOMETRIC_FIELD),
```
j3r3m1, line 165: Shouldn't they made valid before ? We will be back to put makevalid everywhere otherwise no ?

Issue #772, opened 2022-10-12
Acoustic absorption coefficient and ground table


ebocher:
As geoclimate proposes a method to estimate a reference traffic, it'd be nice to offer a way to return the sound absorption coefficient in a ground table. This will facilitate the link with NoiseModelling community (https://github.com/Universite-Gustave-Eiffel/NoiseModelling).
So we can add in the input config file

```json
"parameters": {
        "rsu_indicators": {
            "indicatorUse": [
                "LCZ",
                "TEB",
                "UTRF"
            ],
            "svfSimplified": true,
            "estimateHeight": true
        },
        "noise_indicators": {
            "ground" : true
         }
    }
```
The ground table should contain water, vegetation and imprevious_areas.
The schema of the ground table should be defined with 3 columns : 

-  id_ground : unique identifier
- the_geom : polygon or multipolygon
- g : acoustic absorption coefficient 

@j3r3m1 @gpetit @nicolas-f  @pierromond

ebocher:
Working with the NoiseModelling team offers new research opportunities in the field of "naturally ventilated buildings and noise mapping".

j3r3m1:
If I understand well, the ground table is built according to OSM data and may become a table in the generic model (when "noise_indicators" is set to True) ?

ebocher:
The final configuration file to produce data for noise simulation will be : 

```
"parameters": {
        "rsu_indicators": {
            "indicatorUse": [
                "UTRF"
            ],
        },
        "noise_indicators": {
            "ground" : true
         },
        "road_traffic": true
    }
```

where  

"indicatorUse": [  "UTRF"  ]
is used to estimate the OSM building height

"noise_indicators": { "ground" : true}
is used to set the noise indicators to compute. Currently only the ground table as described above

"road_traffic": true"
is used to set if the road_traffic table is build or not

IMPORTANT : 

GeoClimate does not format the column names for the output building and the road_traffic table as required by NoiseModelling.
Therefore, the NoiseModelling team needs to adapt the column names,
 e.g. replace height_roof with height

ebocher:
Done ;)
I will write a page on the wiki asap

PR #776, opened 2022-10-26
Automatic ground table construction

ebocher:
This PR introduces new algorithms to compute a unified ground layer. A ground layer merged all geometries where the zindex is equal to zero. The ground layer stores all relations between each input layers.
e.g 
It's possible know if  : 
- some part of the road is covered by vegetation
- a road is covered by a building
- a building is covered by vegetation

While the ground layer allows to detect of all overlays, it also permits to build a continuous land cover map by setting priorities when several objects overlap.  Useful to create a  sound absorption coefficient table (#772) or to group  geographical objects according  micro-climate effects.

Input layers
ground_input_layers

Merged layers (find smallest common geometry part)
ground_merged_layers

Tagged smalles geometry according input layers

ground_markerd_layers

ground_building_overlaps

ground_legend

@j3r3m1 @ELSW56   @simogeo @gpetit  @nicolas-f @pierromond


elbeejay:
Hi @ebocher - glad you are continuing to develop `geoclimate`. Not sure why I was tagged in this PR, but to be clear on the JOSS side of things, the JOSS manuscript and review pertain specifically to the version that was associated with the publication. While of course further development of the package is encouraged, we do not need to be alerted of it, and the JOSS paper will not change which archive and software version it references (although the source repository is of course the "live" one, and in Zenodo it will alert a viewer if a newer version is available).

ebocher:
oupps @elbeejay  this is an error on my side ;-(
Nevertheless, thank you for the interesting comment

ebocher:
The ground acoustic table is ready. All overlaping geometries are removed based on a set of priorities.
e.g : building and road surfaces are clipped.

ground_acoustics

ground_acoustics_zoom

Now we need to set the good G value according the land types defined by GeoClimate.

PR #787, opened 2022-11-22
Reorganize module WorkflowGeoIndicators.groovy

j3r3m1:
This PR will reorganize the module WorkflowGeoIndicators in order to simplify the use of individual workflows in the future and to facilitate the use of own RSU by users.


j3r3m1:
Remains two main steps which are:
- replacing part of the code in computeAllGeoIndicators by the corresponding subworkflow
- isolating the part of the code dedicated to building height estimation in a separate workflow.

Then an other PR would be to leave the possibility to skip the spatial unit creation if the user wants to use its own RSU.

j3r3m1:
@ebocher an error is obtained in the test called [runBDTopoWorkflowWithSRID()]
(https://github.com/j3r3m1/geoclimate/blob/6310973d65f6a871bb6aa7b48e9e46f06549f91e/bdtopo_v2/src/test/groovy/org/orbisgis/geoclimate/bdtopo_v2/WorkflowBDTopo_V2Test.groovy#L317). It seems the parameter snappingDistance passed to the workflow computeAllGeoindicators is null [at this stage] (https://github.com/j3r3m1/geoclimate/blob/6310973d65f6a871bb6aa7b48e9e46f06549f91e/bdtopo_v2/src/main/groovy/org/orbisgis/geoclimate/bdtopo_v2/WorkflowBDTopo_V2.groovy#L1326). I would have expected that in that case the default value for snappingDistance in the workflow computeAllGeoindicators() is taken by default (0.01) but it is not the case, the value is kept as null which cause SQL issue afterward.

What would be the best solution to solve that ? As a simple fix I can modify (add) the value of snappingDistance in the test but you might have a more generic way to solve it ?

geoindicators/src/main/groovy/org/orbisgis/geoclimate/geoindicators/WorkflowGeoIndicators.groovy
```diff
@@ -1035,430 +1279,110 @@ IProcess computeAllGeoIndicators() {
...
+                            rsu_lcz            : null,
+                            zone               : zoneTable,
+                            rsu_utrf_area      : null,
+                            rsu_utrf_floor_area: null,
+                            building_utrf      : null,
+                            building           : buildingTableName]
                 }
+
+                //The spatial relation tables RSU and BLOCK  must be filtered to keep only necessary columns
+                def rsuRelationFiltered = prefix prefixName, "RSU_RELATION_"
```
ebocher, line 1327: Not sure to understand this filter

ebocher: rsuRelationFiltered used intead of rsuTable
If yes is rsuTable deleted ?

ebocher (commented):
What a big change ;-)
Just some comments.
By the way, excellent  job.

geoindicators/src/main/groovy/org/orbisgis/geoclimate/geoindicators/WorkflowGeoIndicators.groovy
```diff
@@ -875,6 +875,244 @@ IProcess computeRSUIndicators() {
         }
     }
 }
+
+
+/**
+ * Compute the typology indicators
```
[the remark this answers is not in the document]
j3r3m1: This part of the code does not include anything concerning TEB. computeRSUIndicators is not performed here, only typology classifications (UTRF and LCZ) are computed here (based on geoindicators tables)

geoindicators/src/main/groovy/org/orbisgis/geoclimate/geoindicators/WorkflowGeoIndicators.groovy
j3r3m1: This part of the code was already existing in the previous version. Not sure what contains rsuTable but it might be useless. Yes rsuTable is deleted.

PR #792, opened 2022-12-08
Geoclimate verbose command

ebocher:
Add verbose command to display or not the log trace

Syntax : 

```
-w osm  -f  /tmp/conf.json  verbose
```

Not yet available with the config file


j3r3m1:
Is there a way to get all log information (something like debug) ?

PR #793, opened 2022-12-15
Add the possibility to custom the logger

ebocher:
Now it's possible with the Geoclimate CLI to custom the log messages.

Examples :

```
-w osm -f /tmp/config.json -l off //No messages
```

```
-w osm -f /tmp/config.json -l info //info messages
```

```
-w osm -f /tmp/config.json -l debug //info + debug messages
```

### DIFF
--- a/bdtopo_v2/pom.xml
+++ b/bdtopo_v2/pom.xml
@@ -60,8 +60,8 @@
             <artifactId>slf4j-api</artifactId>
         </dependency>
         <dependency>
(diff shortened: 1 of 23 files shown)


j3r3m1:
By default do you plan to have info instead of off ? Because currently it seems there is no informations logged when we apply the workflow

ebocher:
> By default do you plan to have info instead of off ? Because currently it seems there is no informations logged when we apply the workflow

The default log level is set to info.

Issue #840, opened 2023-09-20
Error calculating indicators in "only sea" areas


j3r3m1:
When GeoClimate is applied on areas where only sea is present, the following error happens. I start to work on that to create a unit test to see how to pass this when necessary.

The OSM workflow has been started.
Please wait...
17:42:59.137 [main] INFO org.orbisgis.geoclimate.osm.OSM - 1 osm areas will be processed
17:43:00.339 [main] INFO org.orbisgis.geoclimate.osmtools.OSMTools - Extract the OSM data
17:43:03.257 [main] INFO org.orbisgis.geoclimate.osmtools.OSMTools - Downloading the OSM data from overpass api in /tmp/f31f5a16798717f0a15182beb8af61c4b8e31bc9ab71fa3142d60a3ed56f8c02.osm
17:43:03.262 [main] INFO org.orbisgis.geoclimate.osmtools.OSMTools - The OSM file has been downloaded at /tmp/f31f5a16798717f0a15182beb8af61c4b8e31bc9ab71fa3142d60a3ed56f8c02.osm.
17:43:03.323 [main] INFO org.orbisgis.geoclimate.osmtools.OSMTools - Load the OSM file in the database.
17:43:03.384 [main] INFO org.orbisgis.geoclimate.osmtools.OSMTools - The input OSM file has been loaded in the database.
17:43:03.533 [main] INFO org.orbisgis.geoclimate.osm.OSM - Building layer created
17:43:03.542 [main] INFO org.orbisgis.geoclimate.osm.OSM - Road layer created
17:43:03.551 [main] INFO org.orbisgis.geoclimate.osm.OSM - Rail layer created
17:43:03.584 [main] INFO org.orbisgis.geoclimate.osm.OSM - Vegetation layer created
17:43:03.602 [main] INFO org.orbisgis.geoclimate.osm.OSM - Water layer created
17:43:03.619 [main] INFO org.orbisgis.geoclimate.osm.OSM - Impervious layer created
17:43:03.636 [main] INFO org.orbisgis.geoclimate.osm.OSM - Urban areas layer created
17:43:03.660 [main] INFO org.orbisgis.geoclimate.osm.OSM - Coastline layer created
17:43:03.776 [main] INFO org.orbisgis.geoclimate.Geoindicators - Create the units of analysis...
17:43:04.211 [main] INFO org.orbisgis.geoclimate.Geoindicators - Start computing the geoindicators...
[...]

PR #841, opened 2023-09-21
Fix issue 840 (workflow stops when only the zone is only sea)

j3r3m1:
Should solve #840


j3r3m1:
@wurtzj: once accepted, the snapshot version of the next day should work

geoindicators/src/main/groovy/org/orbisgis/geoclimate/geoindicators/RsuIndicators.groovy
```diff
@@ -1469,6 +1469,18 @@ String smallestCommunGeometry(JdbcDataSource datasource, String zone, String id_
                                                         MAX(BUILDING) AS BUILDING, MAX(RAIL) AS RAIL, ${id_zone} FROM $allInfoTableName GROUP BY ${ID_COLUMN_NAME}, ${id_zone};
                                       DROP TABLE IF EXISTS ${tablesToMerge.keySet().join(' , ')}, ${allInfoTableName}, ${tmpTablesToDrop.join(",")}""".toString()
         }
+        else{
+            datasource """DROP TABLE IF EXISTS $outputTableName;
+                CREATE TABLE $outputTableName(AREA DOUBLE,
```
ebocher, line 1474: Use double precision

PR #842, opened 2023-09-21
Fix discussion #833

j3r3m1:


ebocher:
Does this change tested on a zone with a retail area ?

PR #846, opened 2023-09-26
Add new exact indicators at grid scale

ebocher:
Here the list 

- [x] ASPECT_RATIO (GeoClimate method https://github.com/orbisgis/geoclimate/wiki/RSU-indicators#aspect_ratio)
- [x] SVF (done with Ground Sky View Factor process)
- [x] HEIGHT_OF_ROUGHNESS_ELEMENTS (Geometric average of building heights (LCZs 1–10) and tree/plant heights (LCZs A–F) (m))
- [x]  TERRAIN_ROUGHNESS_CLASS (Davenport et al.’s (2000) classification of effective terrain roughness (z0 ) for city and country landscapes)

@j3r3m1 @MGousseff

### DIFF
--- a/geoindicators/src/main/groovy/org/orbisgis/geoclimate/geoindicators/WorkflowGeoIndicators.groovy
+++ b/geoindicators/src/main/groovy/org/orbisgis/geoclimate/geoindicators/WorkflowGeoIndicators.groovy
@@ -2062,7 +2062,7 @@ String rasterizeIndicators(JdbcDataSource datasource,
         //Because all other indicators have been yet computed we just alter the table with the aspect_ratio column
         datasource.execute("""
         ALTER TABLE $grid_indicators_table ADD COLUMN ASPECT_RATIO DOUBLE PRECISION;
-        UPDATE $grid_indicators_table SET ASPECT_RATIO = case when building_fraction = 1 then free_external_facade_density else  free_external_facade_density / (1 - building_fraction) end
+        UPDATE $grid_indicators_table SET ASPECT_RATIO = case when building_fraction = 1 then null else  free_external_facade_density / (1 - building_fraction) end
         """.toString())
     }
     tablesToDrop << createScalesRelationsGridBl
--- a/geoindicators/src/main/groovy/org/orbisgis/geoclimate/geoindicators/WorkflowGeoIndicators.groovy
+++ b/geoindicators/src/main/groovy/org/orbisgis/geoclimate/geoindicators/WorkflowGeoIndicators.groovy
@@ -692,7 +692,7 @@ String computeRSUIndicators(JdbcDataSource datasource, String buildingTable,
                             AS SELECT 1-extended_free_facade_fraction AS GROUND_SKY_VIEW_FACTOR, $columnIdRsu 
                             FROM ${computeExtFF}; DROP TABLE ${computeExtFF}""".toString()
         } else {
-            def computeSVF = Geoindicators.RsuIndicators.groundSkyViewFactor(datasource, intermediateJoinTable,
+            def computeSVF = Geoindicators.RsuIndicators.groundSkyViewFactor(datasource, intermediateJoinTable,"id_rsu",
                     buildingTable, parameters.svfPointDensity,
                     parameters.svfRayLength, parameters.svfNumberOfDirection,
                     temporaryPrefName)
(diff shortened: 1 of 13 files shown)


ebocher:
The HEIGHT_OF_ROUGHNESS_ELEMENTS is not yet available because the effectiveTerrainRoughnessLength algorithm must be replaced by frontalAreaIndexDistribution to compute its exact value.

ebocher:
To limit the number of calls to basic geometric functions such as area, I suggest calculating them when the data is formatted.
e.g building layer must contains a column name area, perimeter...

ebocher:
projectedFacadeAreaDistribution and frontalAreaIndexDistribution do not used the same notation for the result columns.

The first one use XX_INDEX_H50_D0_30 and the second one XX_INDEX_H50_${layer_top}_D0_30

See : https://github.com/orbisgis/geoclimate/blob/master/geoindicators/src/main/groovy/org/orbisgis/geoclimate/geoindicators/RsuIndicators.groovy#L1940

j3r3m1:
> The first one use XX_INDEX_H50_D0_30 and the second one XX_INDEX_H50_${layer_top}_D0_30

Seems more relevant to use the second notation: even though it induces longer column names, it is probably simpler to collect informations such as deltaH in the interval in the second case.

ebocher:
Here we are ;-)

geoindicators/src/main/groovy/org/orbisgis/geoclimate/geoindicators/WorkflowGeoIndicators.groovy
```diff
@@ -2058,7 +2062,7 @@ String rasterizeIndicators(JdbcDataSource datasource,
         //Because all other indicators have been yet computed we just alter the table with the aspect_ratio column
         datasource.execute("""
         ALTER TABLE $grid_indicators_table ADD COLUMN ASPECT_RATIO DOUBLE PRECISION;
-        UPDATE $grid_indicators_table SET ASPECT_RATIO = free_external_facade_density / (1 - building_fraction)
+        UPDATE $grid_indicators_table SET ASPECT_RATIO = case when building_fraction = 1 then free_external_facade_density else  free_external_facade_density / (1 - building_fraction) end
```
j3r3m1, line 2065: I do not get it. I suppose when building fraction = 1, then aspect ratio cannot be calculated I would go for null

geoindicators/src/test/groovy/org/orbisgis/geoclimate/geoindicators/RsuIndicatorsTests.groovy
```diff
@@ -104,6 +104,30 @@ class RsuIndicatorsTests {
...
 
+    @Test
+    void freeExternalFacadeDensityExactTest2() {
+        // Only the first 1 first created buildings are selected for the tests
+        h2GIS """
+                DROP TABLE IF EXISTS tempo_build, tempo_rsu; 
+                CREATE TABLE tempo_build(id_build int, the_geom geometry, height_wall double);
+                INSERT INTO tempo_build VALUES (1, 'POLYGON((0 0, 10 0, 10 10, 0 10, 0 0))'::GEOMETRY, 10),
+                                               (2, 'POLYGON ((10 0, 20 0, 20 20, 10 20, 10 0))'::GEOMETRY, 10),
+                                                (3, 'POLYGON ((30 30, 50 30, 50 50, 30 50, 30 30))'::GEOMETRY, 10);
```
j3r3m1, line 115: with this test there is no shared wall between your buildings. You should add a fourth one sharing wall with one of your other buildings.
I propose (4, 'POLYGON ((30 50, 50 50, 50 70, 30 70, 30 50))'::GEOMETRY, 10)

ebocher: Buidling 1 and 2 share a some part

geoindicators/src/main/groovy/org/orbisgis/geoclimate/geoindicators/WorkflowGeoIndicators.groovy
ebocher: oupps.. This is a stupid rule ; I will fix it.

Issue #869, opened 2023-11-08
Trouble with building height estimation


j3r3m1:
There is a bug using OSM on the bounding box [33.8, 7.6, 34, 7.8].

The following error message is logged:
>11:11:02.024 [main] INFO class org.orbisgis.geoclimate.Geoindicators -- No building height to estimate
org.codehaus.groovy.runtime.typehandling.GroovyCastException: Cannot cast object 'false' with class 'java.lang.Boolean' to class 'java.util.Map'
    at org.codehaus.groovy.runtime.typehandling.DefaultTypeTransformation.continueCastOnSAM(DefaultTypeTransformation.java:404)
    at org.codehaus.groovy.runtime.typehandling.DefaultTypeTransformation.continueCastOnNumber(DefaultTypeTransformation.java:315)
    at org.codehaus.groovy.runtime.typehandling.DefaultTypeTransformation.castToType(DefaultTypeTransformation.java:243)
    at org.codehaus.groovy.runtime.ScriptBytecodeAdapter.castToType(ScriptBytecodeAdapter.java:615)
    at org.orbisgis.geoclimate.geoindicators.WorkflowGeoIndicators.estimateBuildingHeight(WorkflowGeoIndicators.groovy:1436)
    at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:104)
    at java.base/java.lang.reflect.Method.invoke(Method.java:577)
    at org.codehaus.groovy.runtime.callsite.PlainObjectMetaMethodSite.doInvoke(PlainObjectMetaMethodSite.java:48)
    at org.codehaus.groovy.runtime.callsite.PogoMetaMethodSite$PogoCachedMethodSiteNoUnwrap.invoke(PogoMetaMethodSite.java:178)
    at org.codehaus.groovy.runtime.callsite.PogoMetaMethodSite.callCurrent(PogoMetaMethodSite.java:57)
    at org.codehaus.groovy.runtime.callsite.CallSiteArray.defaultCallCurrent(CallSiteArray.java:51)
    at org.codehaus.groovy.runtime.callsite.AbstractCallSite.callCurrent(AbstractCallSite.java:171)
    at org.orbisgis.geoclimate.geoindicators.WorkflowGeoIndicators.computeAllGeoIndicators(WorkflowGeoIndicators.groovy:1233)
[...]

PR #870, opened 2023-11-08
Solve building height estimation 

j3r3m1:
When buildings are present but all have a building height already, there is a bug (cf. issue #869). This PR solves it (we might add a unit test, @ebocher what do you think ?)


ebocher (approved):
brilliant !

Issue #881, opened 2023-12-04
"areaTypeAndComposition" should be fullfilled with more types


j3r3m1:
Currently, when we make the GeoClimate calculations, only a small fraction of the buildings are considered in the TypeProportion process (https://github.com/orbisgis/geoclimate/blob/ffa10ae81e498a75f33789f20c14b64aa3dbf167/geoindicators/src/main/groovy/org/orbisgis/geoclimate/geoindicators/GenericIndicators.groovy#L677). The default values considered are given at the following line:
https://github.com/orbisgis/geoclimate/blob/ffa10ae81e498a75f33789f20c14b64aa3dbf167/geoindicators/src/main/groovy/org/orbisgis/geoclimate/geoindicators/WorkflowGeoIndicators.groovy#L372

We should definitely extend this map to at least consider the "house" keyword as a "house" fraction or include it in the "residential" type. We can add a test to check that the percentage of building taken into account is near 100%.

j3r3m1:
@wurtzj

- which building types are used in TEB ? 
- would you be more interested by the building type or the building use (or both) ? Here are the possible type: https://github.com/orbisgis/geoclimate/wiki/Building-layer#type-and-main_use-column-values (we can gather them into main categories)
- do you want the fraction of each type based on the building area or building floor area ?

j3r3m1:
OK, we still have to define what are the "main" types (@wurtzj).
The idea is then to directly create a new attribute at building input scale which will gather some of the existing types into a more generic one.

wurtzj:
Hi @j3r3m1 @ebocher 



What is the difference between building area and building floor area ? walls ? number of levels ?
I would rather say we need the fraction of each use. So if there are commercial activities on first floor, and residential on other floors, then a fraction of each is ideal.

TEB deal with building usage with an approach is similar to a tile approach with flux
aggregation (e.g. using TEB for urban areas and ISBA for rural areas and aggregate fluxes towards
the atmosphere), but the tiles here represent different building uses and human behaviours and the
fluxes towards the building envelope (wall and roof) and the outdoor air are aggregated.

In TEB we use these types :

Fraction of individual housing
building use

Fraction of collective housing
building use

Fraction of commercial
building use

Fraction of tertiary building
use

Fraction of industrial building
use

Fraction of non heated
buildings

Fraction of low-rise building

Fraction of mid-rise building

Fraction of high-rise building

Fraction of activity buildings

Fraction of activity buildings

Fraction of households with
[...]

PR #944, opened 2024-03-14
Modify types

j3r3m1:
Creates new categories for calculating the fraction of each building type according to #881.

### DIFF
--- a/geoindicators/src/test/groovy/org/orbisgis/geoclimate/geoindicators/WorkflowGeoIndicatorsTest.groovy
+++ b/geoindicators/src/test/groovy/org/orbisgis/geoclimate/geoindicators/WorkflowGeoIndicatorsTest.groovy
@@ -85,22 +85,22 @@ class WorkflowGeoIndicatorsTest {
         listFloorBuildTypLcz = []
         listFloorBuildTypTeb = []
         for (type in parameters.buildingAreaTypeAndCompositionTeb.keySet()) {
-            listBuildTypTeb.add("AREA_FRACTION_${type}".toString())
+            listBuildTypTeb.add("AREA_FRACTION_${type}".toString().toUpperCase())
         }
         for (type in parameters.floorAreaTypeAndCompositionTeb.keySet()) {
-            listFloorBuildTypTeb.add("FLOOR_AREA_FRACTION_${type}".toString())
+            listFloorBuildTypTeb.add("FLOOR_AREA_FRACTION_${type}".toString().toUpperCase())
         }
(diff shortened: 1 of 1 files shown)


bdtopo/src/main/groovy/org/orbisgis/geoclimate/bdtopo/InputDataFormatting.groovy
```diff
@@ -97,48 +97,152 @@ String formatBuildingLayer(JdbcDataSource datasource, String building, String zo
                      "Annexe"                            : ["annex": "building"],
                      "Industriel, agricole ou commercial": ["commercial": "commercial"],
                      "Bâtiment"                          : ["building": "building"],
-                     "Industrie lourde"                  : ["heavy_industry": "industrial"]
+                     "Industrie lourde"                  : ["industrial": "industrial"]
```
ebocher, line 100: Why not keep heavy_industry type ?

osm/src/main/resources/org/orbisgis/geoclimate/osm/urbanAreasParams.json
```diff
@@ -45,36 +59,42 @@
...
+      "amenity": [
+        "community_centre"
+      ]
+    },
     "military": {
       "landuse": [
         "military"
       ]
     },
-    "transport": {
```
ebocher, line 72: This will introduce some issue in urban areas because this type will not be pushed when the buildings are formatted. Rigth ?

ebocher (commented):
Great job. However, I'm still having some difficulty imagining the consequences for the urban_areas layer and its use to enrich building types.

bdtopo/src/main/groovy/org/orbisgis/geoclimate/bdtopo/InputDataFormatting.groovy
j3r3m1: If we want to follow the OSM types heavy industry is not defined, it has to be defined later on (as it is the case for the LCZ case)

geoindicators/src/main/groovy/org/orbisgis/geoclimate/geoindicators/RsuIndicators.groovy
```diff
@@ -1086,6 +1094,19 @@ String linearRoadOperations(JdbcDataSource datasource, String rsuTable, String r
     }
 }
 
+String getRoadDirIndic(int d, Integer angleRangeSize, Integer lev){
```
ebocher, line 1097: Doc ?

ebocher:
I'll propose some adjustements in unit test.

PR #949, opened 2024-03-28
Start adding indicators in the chain

j3r3m1:
Add missing indicators in the chain (#704)


j3r3m1:
Note that more informations about the WRF needed indicators can be found at #WRF
 and that the documentation concerning the new indicators can now be found at the RSU page (https://github.com/orbisgis/geoclimate/wiki/RSU-indicators)

PR #953, opened 2024-04-10
Troubles with types

j3r3m1:
Residential were two times in the floorAreaTypeAndCompositionTeb...


j3r3m1:
> When there is no building, all values are set to 0 except the fraction of undefined which is set to 1. All should be set to null in this case in my opinion

@ebocher

PR #963, opened 2024-05-06
Add the recently added building types into default parameters and in …

j3r3m1:
…LCZ calc


j3r3m1:
Seems there was some missing info in the last PR. Sorry for not seeing it directly.

PR #971, opened 2024-05-29
Allow formating GIS layers from the workflow config file

ebocher:
It's now possible to use GeoClimate workflow just to prepare the GIS layers.
Unit tests added


j3r3m1 (approved):
Topissiiiiiiime !

PR #993, opened 2024-09-13
Update ground_acoustic_absorption.json

sachabaclet:
Table VI.1 from CNOSSOS reference. I updated the G coefficients.
image


ebocher:
Thanks a lot !

For legal reasons, contributors are asked to provide a contributor license agreement (CLA).

The message need to include the following statement:

"I wrote the code, it's mine, and I'm contributing it to H2GIS for distribution licensed under the [LGPL 3.0](http://www.gnu.org/copyleft/lgpl.html)."

So if it's ok for you, please add in this PR refer the CLA message.

sachabaclet:
I wrote the code, it's mine, and I'm contributing it to H2GIS for distribution licensed under the [LGPL 3.0](http://www.gnu.org/copyleft/lgpl.html).

ebocher:
Thanks a lot

PR #1008, opened 2024-11-19
Add TARGET model input

ebocher:
This PR adds a new option in the workflow to compute the landcover layer requiered to run the TARGET model (see https://github.com/jixuan-chen/target/tree/main)

The landcover layer is a regular grid saved by geoclimate in flatgeobuffer format as the grid_indicators layer.
By default the grid is computed on 100 x 100 resolution (meters) but the user can set a new resolution using the grid_indicators parameters. See the examples.

Enable the default target configuration

```groovy
"rsu_indicators": ["indicatorUse" : ["TARGET"]  ]
```

Customize the target grid size
```groovy
"rsu_indicators": ["indicatorUse" : ["TARGET"]],
"grid_indicators"   : [
                                "x_size"    : 200,
                                "y_size"    : 200
                        ]
```

@j3r3m1


j3r3m1:
Great ! It would be nice to have the same kind of tutorial than for WRF (https://github.com/orbisgis/geoclimate/wiki/Creates-WRF-inputs-on-Linux)

PR #1017, opened 2024-12-10
Force roof fraction to 0.75

ebocher:
When the building fraction is greater than 0.75 force it to 0.75.

See 

https://github.com/jixuan-chen/target/issues/14

https://github.com/jixuan-chen/target/blob/main/target_py/scripts/Ta_module_new.py#L151C24-L151C28

### DIFF
--- a/geoindicators/src/main/groovy/org/orbisgis/geoclimate/geoindicators/GridIndicators.groovy
+++ b/geoindicators/src/main/groovy/org/orbisgis/geoclimate/geoindicators/GridIndicators.groovy
@@ -386,7 +386,7 @@ String formatGrid4Target(JdbcDataSource datasource, String gridTable) {
                             THE_GEOM,
                             ID_COL, ID_ROW,
                             CAST(row_number() over(ORDER BY ID_ROW DESC) as integer) as "FID",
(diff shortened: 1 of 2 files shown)


geoindicators/src/main/groovy/org/orbisgis/geoclimate/geoindicators/GridIndicators.groovy
```diff
@@ -386,7 +386,7 @@ String formatGrid4Target(JdbcDataSource datasource, String gridTable) {
                             THE_GEOM,
                             ID_COL, ID_ROW,
                             CAST(row_number() over(ORDER BY ID_ROW DESC) as integer) as "FID",
-                            BUILDING_FRACTION AS "roof",
+                            CASE WHEN BUILDING_FRACTION>0.75 THEN 0.75 ELSE END AS "roof",
```
j3r3m1, line 389: ELSE BUILDING_FRACTION sinon ça marche pas =)

ebocher: Misère...

PR #1019, opened 2024-12-19
Add lanes column on road layer

ebocher:
Keep lanes value on road when they are known.

@gpetit @pierromond

Todo : Update wiki


gpetit:
Probably the fastest integration of a need in history ;-)

ebocher:
but with a bug in unit test ;-)
I'm on it

Issue #1070, opened 2025-10-02
Too many building types calculated at grid scale


j3r3m1:
Currently, at grid scale, eahc building type can gives a building type column while they are gathered by categories in the RSU. We will use the same categories in grid scales than as at RSU scale

j3r3m1:
Once this is solved, we could calculate the LCZ directly at grid scale or any other RSU

PR #1071, opened 2025-10-02
Modify function but no test performed

j3r3m1:
Solve #1070


j3r3m1:
Waiting from @ebocher modifications

PR #1115, opened 2026-04-27
Fix spatial units production

ebocher:
Replace ST_Clip by ST_Intersection to compute spatial units


j3r3m1:
Everything's rock for me !

Issue #1123, opened 2026-06-18
~100% Impervious over large areas in Spain (wind turbines ?)


wurtzj:
Hello,

I have identified large areas over Spain that are 100% impervious.
This seems related to wind turbines, but as the openStreetMap seems to have changes compared to the database I use it is difficult to assess exactly.
Obviously under the windturbines the soil in not impervious. 

Thanks for the help :)

Kind Regards,

Jean

<img width="1189" height="809" alt="Image" src="https://github.com/user-attachments/assets/ab1edbb4-b4dc-47e8-9007-d025c8e26583" />

<img width="1079" height="787" alt="Image" src="https://github.com/user-attachments/assets/694919aa-1e4c-486c-8004-562826ae8ae4" />

<img width="1270" height="804" alt="Image" src="https://github.com/user-attachments/assets/2bd58252-51e9-4000-a79f-cce67c1e44e9" />

j3r3m1:
Hi @wurtzj 
Thanks for reporting. Could you give more informations about the area location ?

j3r3m1:
OK I have found it !
https://www.openstreetmap.org/node/8936385122#map=15/42.03576/-0.95431

j3r3m1:
I have run four different versions of the DEODE versions and I cannot get the features you have, neither the values for impervious.

Here is what I get with the current OSM data:
<img width="1630" height="858" alt="Image" src="https://github.com/user-attachments/assets/32325b46-4b9e-4c73-a557-536a87ad74db" />

However, if I use data from 2023-01-01 (I do not remember exactly when was the data you have downloaded), here is what I get with the oldest version of DEODE (version from January 8 - the RSU are quite different if I use the earliest version but the issue remains the same, there is an impervious fraction far from 0...):



As you can see, it is much closer from what you have obtained (with probably a slightly different version and OSM data that might also differ slightly due to the date chosen).
<img width="1630" height="858" alt="Image" src="https://github.com/user-attachments/assets/781da2eb-d71e-4ec3-afc3-a1942fb51888" />

The reason is that there was a quite strange way to define an area covered by wind power plants. Here is the data visualized from the [overpassturbo API](https://overpass-turbo.eu/) executing the following code:
[date:"2023-01-01T00:00:00Z"];
( node({{bbox}}); <; >; );
out meta;
[...]

PR #1128, opened 2026-09-01
Fix #1123

j3r3m1:
Remove wind power plant parks from impervious


worldpoptools/pom.xml
```diff
@@ -4,12 +4,12 @@
     <parent>
         <artifactId>geoclimate-parent</artifactId>
         <groupId>org.orbisgis.geoclimate</groupId>
-        <version>1.0.1-SNAPSHOT</version>
```
j3r3m1: This one should probably not be changed right ?

PR #1133, opened 2026-09-18
Add of a new parameter, "angle", for grid indicators

nathan-marie:
Added a new parameter, "angle" in degrees, to rotate grid


ebocher:
Please give an example with a json config file and post some figures to show the result.
Thanks

nathan-marie:
I updated my pull request message, I still plan to update my pull request to add tests.

j3r3m1:
Do you have a single cell ? Can you add the scale of your figure in order to better understand and check visually that everything seems fine ?

nathan-marie:
Good idea, here are the figures with a scale bar and a single cell :

demo_without_angle_3
With angle set to 45 degrees :
demo_with_ angle_3

The grid is larger when there is a rotation ( 8846-meter side length without rotation , 10972 meter side length with rotation)
 I think this comes from the fact that the rotation in h2gis calculates the envelope twice. But is that a problem?

j3r3m1:
Excellent ! Ideally, it would be nice to add the "Zone" geometry above your grid geometry in order to check that the grid correctly cover the whole zone geometry (Vannes boundaries in your case).
Moreover, I do not understand why you have a single geometry for your grid. With the provided json indicating a 10 m resolution cell, your grid should contain a lot of grid cells !!! (For the example 1 km resolution would be more appropriate)

nathan-marie:
Sorry, I forgot to mention that for the last two figures, I set the rowCol parameter to true with x_size and y_size set to 1. That is why I only have a single geometry.

j3r3m1:
Is adding the zone above the grid is still feasible anyway ?

nathan-marie:
Yes, here it is. I had to restart Geoclimate, I had been writing over the files to run a test with the differents number of cells.

demo_without_angle_4

With angle set to 45 degrees :

demo_with_ angle_4

It's clear that the angled grille could be smaller

j3r3m1:
The question is indeed: do we want the grid being smaller ? 

Because currently the grid seems to be built on the extent of the initial geometry (I have not checked in detail the algo of the new signature of the H2GIS st_makegrid function but it seems that you are rotating the grid built from the geometry extent). I would rather rotate the grid built from the extent of the rotated geometry. It would mean the following steps:
1. rotate the geometry from -angle
2. create the extent 
3. create the grid
4. rotate the grid from angle

What do you think ? @nathan-marie @ebocher

nathan-marie:
Good question, I feel like the grid should be the smallest possible size, but I don't really know why.
The current rotation system is: 
1. create the extent
2. rotate the extent from -angle
3. create a new extent 
4. create all the cells with a rotation angle

nathan-marie:
But if I go back and modify the h2GIS grid creation function, I get this : 
Capture d’écran du 2026-09-21 15-57-23
With the current method : 
Capture d’écran du 2026-09-21 16-03-34
With the two method : 
Capture d’écran du 2026-09-21 16-03-42
With the non-rotating grid and geometry on top : 
Capture d’écran du 2026-09-21 16-05-10
Test_base0 is the basic geometry (POLYGON((0 0, 2 0, 2 2, 0 0)))
Test_base_rotated is the grid with no rotation (ST_MakeGrid('POLYGON((0 0, 2 0, 2 2, 0 0))'::GEOMETRY, 1, 1))
Test 2115 is the grid with a rotation of 45 degree with the method (1.rotate the geometry from -angle, 2.create the extent ...)
Test 2116 is the grid with a rotation of 45 degree with the method (1. create the extent , 2. rotate the extend from -angle, 3. create a new extent ...)

j3r3m1:
Well, I would rather go for 2116 then. @ebocher ?

ebocher:
@j3r3m1 @nathan-marie 
I'm not sure I understand this discussion, but you need to have a pivot on the geometries or the “extend” to create the oriented grid.
Below are the results using the proposed approach without a pivot

image


```sql
-- 1. POlygon de test
DROP TABLE IF EXISTS input_geometries;
CREATE TABLE input_geometries (id INT PRIMARY KEY, the_geom GEOMETRY(POLYGON, 2154));

INSERT INTO input_geometries VALUES (
    1,
    ST_GeomFromText(
        'POLYGON((
            650000 6860000,
            650800 6860000,
            650800 6860200,
            650600 6860200,
            650600 6860600,
            650800 6860600,
            650800 6860800,
            650000 6860800,
            650000 6860000
        ))',
        2154
    )
);

-- 2. Rotation inverse des géometries (-45°), sans pivot
DROP TABLE IF EXISTS geometries_rotated;
CREATE TABLE geometries_rotated AS
SELECT id, ST_ROTATE(the_geom, RADIANS(-45)) AS the_geom
FROM input_geometries;

-- 2. Rotation inverse des géometries (-45°), sans pivot
DROP TABLE IF EXISTS extend_geometries_rotated;
CREATE TABLE extend_geometries_rotated AS
SELECT ST_EXTENT(the_geom) AS the_geom
FROM geometries_rotated;


-- 3. Grille sur l'extend des geometries (memes cellules 100x100)
DROP TABLE IF EXISTS grid_geometries_rotated;
CREATE TABLE grid_geometries_rotated AS
SELECT * FROM ST_MakeGrid('extend_geometries_rotated', 100, 100);


-- 4. Rotation directe (+45°) de la grille
DROP TABLE IF EXISTS grid_45__final_no_pivot;
CREATE TABLE grid_45__final_no_pivot AS
SELECT ST_ROTATE(the_geom, RADIANS(45)) AS the_geom, id
FROM grid_geometries_rotated;


--Export des données
CALL GEOJSONWRITE('/tmp/geometry.geojson', 'input_geometries', TRUE);
CALL GEOJSONWRITE('/tmp/extend_geometries_rotated.geojson', 'extend_geometries_rotated', TRUE);
CALL GEOJSONWRITE('/tmp/grid_45__final_no_pivot.geojson', 'grid_45__final_no_pivot', TRUE);
```

nathan-marie:
If I am not mistaken, the discussion concerns the grid size with rotation.
Sorry, I forgot to specify the pivot creation in the current method :
1.create the extent
2.create the pivot at the center of the extent for all rotations
3.rotate the extent by -angle by the pivot
4.create a new extent
5.create all the cells with a rotation angle by the pivot

ebocher:
With the following script as demo


```sql
DROP TABLE IF EXISTS pivot;
CREATE TABLE pivot AS
SELECT FLOOR(((ST_XMin(ext) + ST_XMax(ext)) / 2) / 100) * 100 AS cx,
       FLOOR(((ST_YMin(ext) + ST_YMax(ext)) / 2) / 100) * 100 AS cy
FROM (SELECT ST_EXTENT(the_geom) AS ext FROM zone) t;

DROP TABLE IF EXISTS zone_rot;
CREATE TABLE zone_rot AS
SELECT z.id, ST_ROTATE(z.the_geom, RADIANS(-45), p.cx, p.cy) AS the_geom
FROM zone z, pivot p;

DROP TABLE IF EXISTS grid;
CREATE TABLE grid AS
SELECT * FROM ST_MakeGrid('zone_rot', 100, 100);

DROP TABLE IF EXISTS grid_45;
CREATE TABLE grid_45 AS
SELECT ST_ROTATE(g.the_geom, RADIANS(45), p.cx, p.cy) AS the_geom, g.id
FROM grid g, pivot p;
```

